PR #27, opened 2015-07-27
Add distPkgFiles config

Necroskillz:
This fixes two issues:
1. when one of the pkgFiles doesn't exists, it created a new file with content of `undefined`
2. added config for distPkgFiles, because I don't want warning for files that are intended to be missing either from source repo, or dist repo. Also it caused an error with git add anyway if the file didn't exist. distPkgFiles default to distFiles if not defined to be backwards compatible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ Options:
     "verbose": false,
     "force": false,
     "pkgFiles": ["package.json"],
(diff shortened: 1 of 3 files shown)


webpro:
Thanks @Necroskillz! Works great; just released v0.1.3.

PR #28, opened 2015-07-29
Update README.md

wordempire:
The `.release.json` example was not valid JSON, as JSON cannot contain undefined values or comments


webpro:
Thanks @wordempire. It's just a representation of the default settings. I've left the comment in, since I want people to know about it :-)

PR #39, opened 2015-10-11
Update README.md

ahmadawais:
Minor typo, you should not use an alias in documentation.


webpro:
Thanks for catching this, @ahmadawais!

ahmadawais:
:)

Issue #58, opened 2016-10-07
Add "pre" to releaseTypes


bermi:
Currently release-it doesn't allow bumping a pre release number.

For example:

Starting version `0.1.0`

```
release-it premajor --dry-run --non-interactive --prereleaseId=rc
[dry-run] bump package.json 1.0.0-rc.0
```

We've increased the major as expected to `1.0.0` and added the `prereleaseId` `rc.0`

If we need to launch another release candidate, **currently there is no way for release-it to increase `rc.0` to `rc.1`**.

Running pre(major/minor/patch) will update the respective semver version, not the `prereleaseId` counter.

```
release-it prepatch --dry-run --non-interactive --prereleaseId=rc
[dry-run] bump package.json 1.0.1-rc.0
```

This could be **solved by adding `pre` to the list of `releaseTypes`** supported by release-it.

```
release-it pre --dry-run --non-interactive --prereleaseId=rc
[dry-run] bump package.json 1.0.0-rc.1
```

@webpro Are there any concerns about including `pre` as legal `releaseTypes`?

webpro:
Released in v2.5.0.

Issue #56, opened 2016-09-05
Allow pre-releases to be created without modifying the base version


StephanBijzitter:
When I push a project with version `1.2.3`, I have configured my CI to automatically pre-release it as `1.2.4-commithash.0`, this is because I use `release-it prepatch`. I would prefer to create the release, without modifying the patch version, resulting in `1.2.3-commithash.0`.

webpro:
Maybe you can create a small script that resolves the desired version, and then run release-it with that.

StephanBijzitter:
True, but it'd still be nice to have it included in this library :-)

msegado:
The core issue here is that node-semver (and by extension, release-it) only provides an API to _increment_ the version, but `1.2.3-commithash.0` is a version _decrement_ compared to `1.2.3`. There's an undocumented `pre` command in node-semver that does exactly what you want, but per my comment in #59 its probably not safe to expose or document it in release-it (at least not without pinning the exact patch version of node-semver in `package.json`) since it could hypothetically vanish without notice...

StephanBijzitter:
Well, although you are right that it in fact decrements the version, this is only the case when the `1.2.3` version already exists. In our project, this is how we work:

`master` is stable, and as such any changes to any module or additions of new modules, must be tested before merging to master. To do this, we work from branches where every commit is released using release-it under a beta flag. What happens now is the following, assuming an example for a new module that is yet to be released: `flying-dutchman`.

* Developers checkouts out `master`
* Developers creates new branch `module-flying-dutchman`
* Developers creates the module and gives it version `1.0.0`
* Developer pushes branch to origin
* CI releases a beta of this module as `1.0.1-commithash.0`
* Tester tests the beta release in a testing environment
* Developer 2 merges pull request into master
* CI detects change in master and releases this module as `1.0.0`

In other words, the beta release of 1.0.1 was released before the actual release of 1.0.0, which is a bit odd.

PR #59, opened 2016-10-07
Add pre to releaseTypes #58

marianojabdala:


jhartman86:
@webpro i'm running into an issue with this as well. anything I can do to help get this released to NPM?

msegado:
I believe the exposed command should actually be `prerelease` rather than `pre`, as described in the `node-semver` README: https://github.com/npm/node-semver#functions

There is [an internal `pre` command](https://github.com/npm/node-semver/blob/d21444a0658224b152ce54965d02dbe0856afb84/semver.js#L449), but it's not part of their documented public api so it's probably not a good idea to use it (though amusingly it looks like it exactly satisfies @StephanBijzitter's use case over in #56).

(Also, it looks like there was a typo in the docs that references a non-existent "`prelease`" command; might be a good idea to delete it if/when you switch from `pre` to `prerelease` to avoid confusion...)

StephanBijzitter:
As for the documentation in the `pre` command linked above, this indeed is exactly what I would like to use. :-)

webpro:
Released in v2.5.0.

PR #69, opened 2016-12-12
Only determine revision range when needed

szdavid92:
Determining the changelog revision can fail, and it is not needed only if the user added the template.

This pull request modifies the behavior of getChangelog() so it only tries to determine the revisions if the changelogCommand contains the template.

### DIFF
--- a/lib/git.js
+++ b/lib/git.js
@@ -174,23 +174,31 @@ function getGithubToken(tokenRef) {
 }
 
 function getChangelog(options) {
-    if(options.changelogCommand) {
-        var prevVersion = config.process.get('prevVersion') || options.prevVersion;
-        var prevTag = util.format(options.tagName, prevVersion);
-        return tagExists(prevTag).then(function(hasTag) {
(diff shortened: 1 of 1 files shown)


lib/git.js
```diff
@@ -174,23 +174,31 @@ function getGithubToken(tokenRef) {
...
+    })
+  }
+
+  if(options.changelogCommand) {
+    if(options.changelogCommand.match(/\[REV_RANGE\]/) {
+      var prevVersion = config.process.get('prevVersion') || options.prevVersion;
+      var prevTag = util.format(options.tagName, prevVersion);
+      return tagExists(prevTag).then(function(hasTag) {
+          var command = options.changelogCommand.replace(/\[REV_RANGE\]/, hasTag ? prevTag + '...HEAD' : '');
+          return runCommand(command);
```
webpro: Shouldn't this be a call to `runChangelogCommand` instead (also below)?

szdavid92: Good catch! Fixed.

webpro:
Thanks @szdavid92, I've released this in [v2.5.2](https://github.com/webpro/release-it/releases/tag/2.5.2).

PR #70, opened 2016-12-12
Parse --increment option

szdavid92:
We should try to parse the increment option as well


webpro:
This already works (e.g. `release-it --increment major`), so I don't think this is necessary (the minimist pkg supports aliases and they are used). Let me know if I misunderstood.

szdavid92:
Sorry, this indeed works.

PR #75, opened 2016-12-20
Switch for Enterprise and GitHub host in GitHub API

### LINKED ISSUE #74: Support for GitHub Enterprise (Take 2)
While trying to release to a GitHub Enterprise host, the create release task fails, since the GitHub API Client has a hard coded host (api.github.com), while Enterprise requires host/api/v3

leemi:
If site is github.com, use node-github default (api.github.com).
If site is other (ie. Enterprise) use repo.host.

This closes #74

### DIFF
--- a/lib/git.js
+++ b/lib/git.js
@@ -146,13 +146,14 @@ function pushTags(version, pushUrl) {
     });
 }
 
-function initGithubClient() {
+function initGithubClient(repo) {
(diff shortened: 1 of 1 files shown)


webpro:
I've used your commit and released it in [v2.5.3](https://github.com/webpro/release-it/releases/tag/2.5.3). Thanks @leemi!

PR #77, opened 2016-12-23
add option to disable commiting and pushing

szdavid92:
Added an option
```
noCommit
```
which is when set to a truthy value, disables the `commit` and the `push` tasks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,7 @@ Options:
     "pkgFiles": ["package.json"],
     "increment": "patch",
     "prereleaseId": null,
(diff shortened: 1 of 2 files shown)


webpro:
Thanks @szdavid92, I need a bit more time to dive into this, but I'll get back to this soon.

webpro:
Thanks for this PR, but I've just [released](https://github.com/webpro/release-it/releases) the first beta of v3, which includes this feature. Feel free to check it out and I'm happy to hear any feedback.

PR #83, opened 2017-02-09
Add releaseDate option

tiste:
It will add a `releaseDate` option. If set to `true`, the `package.json` will be added a `releaseDate` to the current date.


webpro:
Hi @tiste, thanks for your PR. But why add this field to package.json? It's not a [standard field](https://docs.npmjs.com/files/package.json). (You do get `time` info with the `npm info [pkg]` command.) You can use the `buildCommand` or `src. beforeStageCommand` option for this.

PR #101, opened 2017-06-25
Add github.draft (close #100)

### LINKED ISSUE #100: Github release as draft
It would be neat to have an option to make a Github release as a draft instead of a published release.

That would allow to edit the release before the actual publish. It also help when you setup release-it the first time and don't want to screw things up.

MichaelMure:
That was indeed very easy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,7 @@ Here is the full list of settings:
     "release": false,
     "releaseName": "Release %s",
     "preRelease": false,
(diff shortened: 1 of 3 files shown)


webpro:
Thanks, @MichaelMure!

webpro:
This is released in [v2.8.1](https://github.com/webpro/release-it/releases/tag/2.8.1).

PR #151, opened 2017-11-14
Allow options when using release-it as Node.js module

marcuspoehls:
Hey Lars,

I’m using `release-it` as part of a personal CLI to deploy a project. The deploy steps include test runs and actual remote command exeution. I use `release-it` in non-interactive mode to bump the version, commit git changes, tag git release and push the changes.

This PR adds an `options` object as a parameter to `tasks`. It allows users to customize the default behavior of `release-it` when using it as a Node.js module and not from the command line.

I appreciate any feedback if it doesn’t meet your intentions with `release-it` 😃 
Thank you!

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -3,14 +3,14 @@ import { config } from './config';
 import runTasks from './tasks';
 import track, { trackException } from './metrics';
 
-export default function release() {
+export default function release(options) {
   if (config.isShowVersion) {
     version();
   } else if (config.isShowHelp) {
     help();
   } else {
     track('start');
-    return runTasks().then(() => track('end'), trackException);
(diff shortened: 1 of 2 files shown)


webpro:
Thanks for the PR, @marcuspoehls! When I looked into it, I realized that I've been bitten by this challenge before: the `config` _instance_ of `Config` is used in multiple modules. This means that only the `config` in e.g. `tasks.js` would use a fresh configuration instance with your passed `options` (but the other modules would not).

I've solved it in a different way, with no changes necessary on your side in [v4.4.0](https://github.com/webpro/release-it/releases/tag/4.4.0).

marcuspoehls:
Thank you!!

PR #154, opened 2017-11-15
release-it-149 Are projects with submodules contemplated?

vchagaev:
Remove overkill check

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 .tmp*
 coverage
 package-lock.json
+node_modules
--- a/lib/config.js
+++ b/lib/config.js
@@ -1,4 +1,5 @@
 import path from 'path';
+import fs from 'fs';
 import parseArgs from 'yargs-parser';
 import _ from 'lodash';
 import isCI from 'is-ci';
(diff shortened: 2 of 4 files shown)


webpro:
Thanks, @vchagaev!

For default project configs, this might be OK. However, as mentioned in the other issue, I'm slightly concerned that one might release-it from a subdir of a Git project. Because, then:

1. There will not be a `package.json`, and it will fail only _after_ the "Ready to release?" question. Also, the `version` from `package.json` won't be found as a fallback in case no Git tag was found.
2. The `.release-it.json` config file from the root of the project won't be found (resulting in missed config without warning).
3. Build commands might fail as they will be executed from a different dir.

This is why I rather "fail early", then have the user find out about these later on.

Still, the first two can be fixed by looking up those files in ancestor dirs. Then we might be onto something. I understand you don't need this in your case, but are you still interested in handling this?

vchagaev:
What do you think about an extra check before git check. Files .release-it.json and package.json are required and they have to be in the current working dir. This means that user really know what he is doing.

webpro:
That's a great idea. I think it's enough to check for `package.json` only, as `.release-it.json` is optional.

vchagaev:
I can not imagine real project without config file

webpro:
> I can not imagine real project without config file

I think plenty of projects use this without a config file. You can set any configuration from the CLI using arguments.

As to closing this PR: my apologies to be so harsh, but I've incorporated your idea in acdc65bed2a9327667712bee41a5d27b6d1d2e9f, which is now in [5.0.0-beta.0](https://github.com/webpro/release-it/releases/tag/5.0.0-beta.0). I hope you're OK with this!

vchagaev:
Ok

PR #159, opened 2017-11-17
Fix default options overriding

miltador:
According to lodash docs (https://lodash.com/docs/#defaultsDeep) defaults should go as a last argument to _.defaultsDeep() function.


webpro:
Thanks, @miltador. That's correct. Although in practice the related options will not override one another, your PR puts them in the right order.

PR #169, opened 2017-12-06
adjust path to stage directory

owcd:
After upgrading from an older version of the release-it tool I noticed the dist files are no longer being copied to dist repo. This happens because files are copied relative to dist.baseDir. stageDir in the default configuration is part of baseDir parent directory, which means the path needs to be adjusted.

This works with the assumption that baseDir is always a direct child of the root directory.

### DIFF
--- a/lib/tasks.js
+++ b/lib/tasks.js
@@ -1,3 +1,4 @@
+const path = require('path');
 const repoPathParse = require('parse-repo');
 const _ = require('lodash');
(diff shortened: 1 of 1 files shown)


webpro:
Thanks for the PR, @owcd! I've implemented a slightly different solution that `path.resolve` the `dist.stageDir` from the current working dir.

owcd:
Thanks

PR #170, opened 2017-12-06
fix interactive dist release tasks

owcd:
The process of releasing with a distribution repository fails because tasks are passed as an array.


webpro:
Thanks for the catch **and** the fix, @owcd!

PR #171, opened 2017-12-06
make dist repo bump to the same version

owcd:
The version of the dist repo is no longer in sync with the src repo's version.


webpro:
Thanks for catching this, @owcd. Awesome that you've fixed it as well!

PR #180, opened 2017-12-14
Follow config for GH releases and NPM publish

ngotchac:
Don't ask for publishing to NPM or release to Github if the configuration says it shouldn't.
Also returns some info from the release (currently `tagName`, `version`, `changelog`).


webpro:
Hi, @ngotchac! Thanks for your PR. As I prefer small and focused pull requests, I'd like to discuss your original issue here (skip prompts by configuration).

The configurations for the "interactive" vs. "non-interactive" mode are clearly separated (I hope), and I'm not sure why we should "combine" them for just two of the prompts/tasks.

I'm not saying it's a bad idea, but then we should do it for all tasks, right? Perhaps then it would be clean to pass e.g. `npm.publish` to each `prompt` call (and return a noop if it's `false`, just like the `spinner` function).

ngotchac:
Hey, yeah sorry for the mixed PR, I'll make two of them.
So my use-case is that I still want to prompt the user for the tag name, whether a Push needs to be done, etc. But not ask for an npm publish nor for a Github release (because this will be done afterward manually in my case). Would that make sense?
I was just surprised that the prompt would ask me for a GH release when I specified no release in the config, and no Github token has been passed. 
I agree that it could be followed by each step. Let me take another look.

webpro:
Documentation currently says "By default, release-it is interactive and allows you to confirm each task before execution" (https://github.com/webpro/release-it#-interactive-vs-non-interactive-mode). If we merge your ideas, then we can re-state this section.

Maybe open a new PR? :-)

ngotchac:
Closing for #181

PR #181, opened 2017-12-15
Configurable prompts

ngotchac:
Follow-up of #180 

The prompts should follow the options passed to the tasks runner.
@webpro Fell free to edit the README modifications, not sure how to best explain this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 .tmp*
 coverage
 package-lock.json
+node_modules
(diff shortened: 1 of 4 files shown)


webpro:
Thanks @ngotchac, looks good. Could you please point this PR to the `v6` branch instead of the `master` branch? This is quite a change, so I'd prefer to have it in the next major release (even though it's not necessarily a breaking change).

ngotchac:
Sure !

PR #182, opened 2017-12-15
Return results from the tasks runner

ngotchac:
Return some basic (for now) information from the Tasks Runner: `changelog`, `version`.

The enables `release-it` to be used as a library, and results such as the `version` or the `changelog` can be useful for a more complex build process.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 .tmp*
 coverage
 package-lock.json
+node_modules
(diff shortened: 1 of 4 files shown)


webpro:
Thanks @ngotchac, this is a useful PR as well. I've published this (together with #181) in [6.0.0-alpha.1](https://github.com/webpro/release-it/releases/tag/6.0.0-alpha.1).

I've slightly changed the resolved object to contain "parsed" values (not things that are already known configuration). What do you think?

ngotchac:
Looks great, thanks ! :) I can close this PR

PR #205, opened 2018-01-19
Fix typo

prakashregmi:
**Motivation:**
Upon going through  the installation process from documentation, I came to find there was `.` instead of `,` . This PR fixes that typo.


webpro:
Thanks, @prakashregmi!

PR #276, opened 2018-04-08
changed from ssh to https

IamTheHttp:
I updated the docs to work in a "copy paste" manner, the ssh one required a key, while the https one does not.


webpro:
Thanks, @IamTheHttp!

PR #277, opened 2018-04-08
Feature/dynamic errors

IamTheHttp:
This PR will allow users to configure custom error messages, instead the default ones.
This also adds tests to the "tasks.js" file, as I couldn't find any.

We had a need to display custom error messages, so I thought this PR couldn't hurt.


webpro:
Hi @IamTheHttp, thanks for your PR. I'm not sure about the custom error messages, is this common in tools like release-it? Any other examples?

In any case, the tests for tasks.js are obviously a very good idea so I'm definitely interested in that part.

IamTheHttp:
We had issues with the current errors not being descriptive enough for all levels of developers, some were struggling with things like "no upstream" or "working directory must be clean", we wanted to change it to something more helpful such as "please add and commit before releasing" (instead of working dir..)

I didn't want to assume that our error message was correct enough to fit everyone, so I thought a configuration option/escape hatch would be useful. 

Tools similar to release-it do not document any configuration for errors, but they also don't document the errors themselves most of the times.

webpro:
Well, maybe we can try and let everyone benefit from descriptive and clear error messages? I'm definitely interested in improving them where possible.

webpro:
Thanks @IamTheHttp! Great idea, and I've merged your PR. I took your idea a step further, I hope you like it! Let me know if you have any feedback.

PR #291, opened 2018-04-23
Added timeout to config

warnockf:


webpro:
Thanks @warnockf! Could you please change the default to `0`? And put the `timeout` setting within the `github` section of the config? Thanks!

PR #314, opened 2018-06-11
Update package.json engines field

bimbiltu:
async/await is only supported in node >= 7.5 but I decided to just go with the latest LTS since node 7 is EOL very soon


webpro:
Hi, @bimbiltu. With [babel-register](https://babeljs.io/docs/usage/babel-register/), release-it also works with Node v6:

https://github.com/webpro/release-it/blob/0a18cdc96b785e93bf17dcdcb32cc7ef1f67ce00/bin/release-it.js#L7-L11).

If you encounter any issues in Node v6, please let me know.

bimbiltu:
Hmm interesting, after running `yarn add release-it`, running `yarn release-it` works on node 6 however using the programmatic API like this `node -e 'const releaseit = require("release-it")'` will fail

webpro:
Ah, yes. From the command line, the `bin` from package.json is used, and programmatically, the value of the `main` field is the point of entry:

https://github.com/webpro/release-it/blob/0a18cdc96b785e93bf17dcdcb32cc7ef1f67ce00/package.json#L29-L30

That's why release-it can't be `require`-d in Node v6. As this is essentially a stand-alone command-line tool, assuming not many people are using this programmatically in Node v6, and the fact there's no way to make this distinction in package.json (or some other way..?), I'm OK with the status quo. The way it see it, it's more like a bonus that release-it can still be used from CLI in Node v6.

bimbiltu:
@webpro Thanks for the info 👍 , I'm fine with the status quo as well as I usually use node 8/10 these days. This was just something that I had noticed

Issue #333, opened 2018-07-31
Dist repo defaults to not tagging, even if config set to true


nickgravelyn:
I've been using release-it for only a short period and it's working well but I recently started trying to setup a distribution repository and have run into a weird behavior where the default is to not tag the distribution repository. That seems odd given the [default configuration](https://github.com/webpro/release-it/blob/master/conf/release-it.json) has `dist.tag: true`, but even if I set `dist.tag: true` in my own JSON config when I'm going through the steps in the CLI the default is still to not tag that repository.

Is this a misunderstanding on my part or a bug? I really would like my distribution repository to have tags that match my versioning.

EDIT: I also noticed it defaults Github Release to "N" even though my config has it set to true. Are the config values not the defaults in the interactive mode? Are the config values correctly used in the CI mode?

nickgravelyn:
Just realized there is a separate section for prompt settings: https://github.com/webpro/release-it/blob/master/conf/release-it.json#L70. Changing those in my config fixed the issue.

webpro:
Hi, @nickgravelyn! Glad you figured it out. It's documented at https://github.com/webpro/release-it#-interactive-vs-non-interactive-mode, but I'm all ears if this can be improved upon, as I understand the distinction isn't always clear.

rodeyseijkens:
There is still a in my opinion a weird check here cause it only tags the `dist` when the `tagName` doesn't match see: https://github.com/webpro/release-it/blob/master/lib/tasks.js#L196

It took me an hour to figure out why it wasn't tagging properly until I looked into the source and saw it does a specific check after if the `tagName` doesn't match. 

I somewhat get it why the check is there to prevent giving an error when you tag it twice on two different branches. But I have the `src.tag: false` in my config.

Maybe change that line to:
```js
const shouldTag = (dist.tag && !isSameRepo) || (isSameRepo && tagName !== src.tagName) || (dist.tag && !src.tag);
```

webpro:
Closing this as #342 is merged.

PR #342, opened 2018-08-15
Fixed dist.tag not working properly

rodeyseijkens:
Fixed dist.tag not working when src.tag is disabled.
See issue #333


webpro:
Thanks, @rodeyseijkens! Released in [v7.5.1](https://github.com/webpro/release-it/releases/tag/7.5.1)

PR #374, opened 2018-10-17
Add documentation for NPM OTP usage

lgraubner:
I had to look through the issues for this information, might be useful to have in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,14 @@ minified scripts, documentation), provide one or more glob patterns for the `git
 }
 ```
 
+## 🔒 NPM Two-Factor Authentication (OTP)
+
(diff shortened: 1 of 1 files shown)


webpro:
Hi, @lgraubner! Very useful to add indeed. Please use lower-cased "npm". Also, the release-it process doesn't fail if the OTP isn't provided, it asks for it, so that should be reworded a bit. Thanks!

webpro:
I've just tried again, and it WOMM™... Also through an npm script.

Did you run it in non-interactive mode (`-n`)?

wilzbach:
I ran into this with -n, but it would be great if this worked with -n as I don't want to reanswer settings that stay the same all the time.

webpro:
As [2FA/OTP has been added to the readme](https://github.com/webpro/release-it#two-factor-authentication), I'm closing this issue. Feel free to re-open or open a new ticket to discuss this.

PR #379, opened 2018-10-26
Remove origin from --set-upstream when there is not

dani8art:
Hello,

When I was trying to release my `canary` releases which are based on my branch `develop`, I got this error: 

```
$ git push --follow-tags  git@github.com:dani8art/hygeia-js.git --set-upstream origin develop
fatal: refs/remotes/origin/HEAD cannot be resolved to branch.
ERROR fatal: refs/remotes/origin/HEAD cannot be resolved to branch.

error An unexpected error occurred: "Command failed.
Exit code: 1
Command: sh
Arguments: -c release-it --verbose --preRelease=canary
Directory: /root/project
Output:
".
info If you think this is a bug, please open a bug report with the information provided in "/root/project/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
Exited with code 1
```

After some investigations, I realize that it is caused by this line. This changes could be useful for you.

Thanks,
Daniel.

### DIFF
--- a/lib/git.js
+++ b/lib/git.js
@@ -87,7 +87,7 @@ const getLatestTag = () =>
   );
 
 const push = async ({ pushRepo = '', hasUpstreamBranch, args = '' } = {}) => {
-  const setUpstream = hasUpstreamBranch === false ? `--set-upstream origin/${await getBranchName()}` : '';
+  const setUpstream = hasUpstreamBranch === false ? `--set-upstream ${await getBranchName()}` : '';
   return run(`git push --follow-tags ${args} ${pushRepo} ${setUpstream}`);
 };
 
--- a/lib/git.js
+++ b/lib/git.js
@@ -87,7 +87,7 @@ const getLatestTag = () =>
   );
 
 const push = async ({ pushRepo = '', hasUpstreamBranch, args = '' } = {}) => {
-  const setUpstream = hasUpstreamBranch === false ? `--set-upstream origin ${await getBranchName()}` : '';
+  const setUpstream = hasUpstreamBranch === false ? `--set-upstream origin/${await getBranchName()}` : '';
   return run(`git push --follow-tags ${args} ${pushRepo} ${setUpstream}`);
 };
 


webpro:
Hi, @dani8art. Do you have any reference(s) to omit the remote name in Git documentation? From a quick search, this doesn't seem to be a regular way of setting the argument value.

Additionally, the `--set-upstream` feature seems deprecated, so I think we should look into alternative ways of doing things.

dani8art:
Yes, you are right, the `--set-upstream` feature seems to will be deprecated.

After, fight a bit with `git`, I could presume that arguments before `--set-upstream` in `push` command seem to be the remote list of branches. Maybe it is caused because of `release-it` is using the remote URL `git@github.com:dani8art/hygeia-js.git`.

Anyway, as you said we need to look for another solution. The problem is that when I was trying to execute with the default command `git push --follow-tags  git@github.com:dani8art/hygeia-js.git`, I got this:

```
fatal: No upstream configured for branch 'develop'
ERROR No upstream configured for the current branch.
Please either set an upstream branch, or use the `requireUpstream: false` option to have this set this by release-it.
error An unexpected error occurred: "Command failed.
```

Then, I configured my upstream before calling `release-it` command and I got this other error:

```
ERROR To git@github.com:dani8art/hygeia-js.git
   9193ac9..e6093eb  develop -> develop
 * [new tag]         0.0.5-canary.5 -> 0.0.5-canary.5
 ! [rejected]        master -> master (non-fast-forward)
error: failed to push some refs to 'git@github.com:dani8art/hygeia-js.git'
hint: Updates were rejected because a pushed branch tip is behind its remote
hint: counterpart. Check out this branch and integrate the remote changes
hint: (e.g. 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
```

It tried to push all branch, I don't know why?

Regards.

webpro:
```
ERROR To git@github.com:dani8art/hygeia-js.git
   9193ac9..e6093eb  develop -> develop
 * [new tag]         0.0.5-canary.5 -> 0.0.5-canary.5
 ! [rejected]        master -> master (non-fast-forward)
```

Shouldn't that be more like `[branch] -> origin/[branch]`?

webpro:
I've just released [7.6.3-beta.0](https://github.com/webpro/release-it/releases/tag/7.6.3-beta.0) that should fix the issue. Could you please verify this works for you as well? Thanks in advance!

dani8art:
Ok cool!, I'll try and let you know whether it works well for me ;)

plondon:
This fixed my issue, thank you @dani8art for reporting and @webpro for fixing!

Issue #376, opened 2018-10-18
Integrate Custom Changelog w/ Commit


Tornquist:
## Summary

I would like the ability to interact with the changelog being generated, and then append that to the bottom of both the tag, and the commit message.

## Goal

1. Start Release
2. Changelog is generated
3. Changelog is written to tmp file
4. tmp file is opened in `$DEFAULT_EDITOR`
5. tmp file is saved/editor is closed
6. Release continues

## Changes

* Add flag to optionally interact with changelog
* Add additional indicator on `tagAnnotation` and `commitMessage` for additional lines

## Example

tmp file created

```
Merge commit 1
Feature commit A
Feature commit B
Feature commit C
Merge commit 2
Feature commit D
Feature commit E
```

tmp file saved

```
Merge 1 human description
Merge 2 human description
```

Commit Created:

```
<commitMessage>
<changelog>
```

Tag Created:

```
<tagAnnotation>
<changelog>
```

## Notes

Much of my professional work is on other git hosting services. Because of this, the release system that `release-it` is based around is insufficient for my needs. I store changelogs and descriptions in the commit tag/message when I'm handling releases. There is no single release view, so the tag actually holding the descriptions is super important.

My tag commits are formatted similarly to your defaults (with the additional of more lines).

Is this type of extension something you see as fitting into the vision of `release-it`? If so, I'm happy to build it/help build it. I don't want to frame it all in if you don't want those features included in the product though.

PalatinCoder:
Hi @Tornquist @webpro , as I would like to have that myself I started working on it in #380 :tada:

Tornquist:
Awesome! I haven't started yet. I have a shell script that does it, but nothing regarding integration here.

webpro:
Thanks guys!

PR #380, opened 2018-10-30
[WIP] Tag Annotation with Custom Changelog

PalatinCoder:
Hi there 😄 
I started working on the feature requested in #376. So far it seems to have only little impact. Main thing is implemented with the first simple commit. There are still some Todos however, which I hope I can do during the weekend.

## Todo

- [ ] Get it working in PowerShell (so far only bash)
- [ ] Add tests
- [ ] Documentation
- [ ] Ability to edit the message via temp file
- [x] Think about interactive vs. non-interactive
- [ ] same thing for the commit

Also @webpro, where do you think would be a good place to add this to the docs (readme) and how could this be tested well?

Cheers

### DIFF
--- a/conf/release-it.json
+++ b/conf/release-it.json
@@ -19,6 +19,7 @@
     "tag": true,
     "tagName": "%s",
     "tagAnnotation": "Release %s",
+    "tagAnnotationChangelog": false,
     "push": true,
     "pushArgs": "",
     "pushRepo": "origin",
--- a/lib/tasks.js
+++ b/lib/tasks.js
@@ -67,6 +67,7 @@ module.exports = async options => {
       commitArgs,
       tagName,
       tagAnnotation,
+      tagAnnotationChangelog,
       afterReleaseCommand,
       addUntrackedFiles
     } = src;
@@ -147,7 +148,7 @@ module.exports = async options => {
 
     const status = () => git.status();
     const commit = () => git.commit({ path: '.', message, version, args: commitArgs });
-    const tag = () => git.tag(version, tagName, tagAnnotation);
+    const tag = () => git.tag(version, tagName, tagAnnotationChangelog ? tagAnnotation + '\n\n' + changelog : tagAnnotation);
     const push = () =>
       git.push({
         pushRepo: remoteUrl,
@@ -199,7 +200,7 @@ module.exports = async options => {
       _.defaults(npm, options.npm);
 
       const commit = () => git.commit({ path: '.', message, version, args: commitArgs });
-      const tag = () => git.tag(version, tagName, tagAnnotation);
+      const tag = () => git.tag(version, tagName, tagAnnotationChangelog ? tagAnnotation + '\n\n' + changelog : tagAnnotation);
       const push = () => git.push({ args: dist.pushArgs });
       const release = () =>
         githubClient.release({


Tornquist:
Regarding your current commit, I think you should do that for both tag and commits.

As far as testing is concerned, I was kind of at a loss as well. You almost need to wrap it in a shell and actually execute commands against it. ie: open default editor, modify, and trigger a save. That's hard outside of a VM though. It's crazy environmentally dependent.

webpro:
1. Regarding the `tagAnnotation`. There's no need to come up with a new configuration option if we'd use something like this:

```
"tagName": "%s",
"tagAnnotation": "Release ${version}\n\n${changelog}"
```

And use the `template` helper like this (not tested):

```
const tag = () => git.tag(version, tagName, util.template(tagAnnotation, options));
```

This would be a breaking change (for anyone that has overridden the default), as `%s` isn't replaced anymore. I'd be OK with that, and phase out `%s` altogether (and replace with `${version}`). Could be easily backwards compatible by replacing `%s` with `${version}` beforehand.

2. Not sure how to get the interaction working (other than using the `EDITOR` environment variable to get the default editor).

3. Non-interactive means that the interaction with the changelog would be unavailable/disabled.

PalatinCoder:
Thanks @webpro for your feedback!
My intention with introducing a new config option and having it default to the "old" behavior was indeed to not make this a breaking change. However I do like your approach better as it gives more flexibility. I'll try to implement it in a backwards compatible manner, as you suggested.
Once it works for the tag I will also copy it to the commit part, as @Tornquist suggested.
Hardest part is probably to get it PowerShell-compatible 😄 (has a problem with new lines being in the command)

Regarding the temp file, I think the easiest solution would be to just pass `--edit` to git. It will open the default editor by itself and wait for it to close. In non-interactive mode, we would just _not_ add that argument to the command line. This also applies to both commit and tag.

webpro:
If there would be any chance to get/read the modified changelog (either from a return value or by parsing the latest commit), that could be re-used.

Tornquist:
That's easy if you use a tmp file. To make it system independent, you would probably need to do it in the working directory (instead of relying on `/tmp` or something like that).

You'd want to make sure it doesn't get picked up by any of the commit commands.

Alternatively, because you know that the changelog is lines 1...n, you could parse the most recent commit message directly (if a commit was made)

PalatinCoder:
At first, sorry for the long silence guys. Unfortunately I didn't have much time during the last weeks to continue working on this PR, but I did have some time to think 😉 

Regarding the use of the template helper, that actually works great and by not removing the call to `format`it's also perfectly backwards compatible. @webpro do you mind me reverting the first commit and pushing the new implementation, or would you prefer a new PR for the sake of a cleaner commit history? 😃 

Regarding the editing: Using `--edit` flag for git was the easiest solution from release-it's point of view, as we wouldn't have had to worry about which editor to use and other platform dependent stuff. But yes indeed, you would have to edit the changelog twice. 
Using a temp file, we would need to somehow determine which editor to use, ask the user to set the EDITOR env variable for us, or introduce a config option to set the editor.
Finally, editing the commit message and parsing the changelog from it for the tag annotation IMHO just doesn't feel right.. 
In general, I feel like including the changelog in the annotation is release-it's task, while generating the changelog itself isn't. WDYT?
I think a good solution (if you're on linux) would be to use `vipe` from the moreutils in the changelog command. You would just pipe the normal `git log` or whatever you use to generate your changelog into vipe, and it will launch an editor for you (by using a temp file). When you're finished editing, vipe will just spit out the temp file on stdout and release-it can pick it up. I think that would meet @Tornquist 's requirements, while also release-it wouldn't have to do "the hard work" and just include the changelog in the annotation.

And another thing: I'm working on windows primarily and here shelljs is struggling with executing the commands, due to all the new lines from the changelog in it. There is actually an issue about that: shelljs/shelljs#175 
I have no real solution for that yet, other than using git bash or some WSL bash. WDYT about that? Should we just live with it and put a note in the documentation?

webpro:
@PalatinCoder @Tornquist While definitely not a replacement for manually modifying the generated changelog afterwards, but [auto-changelog](https://github.com/CookPete/auto-changelog) allows more flexibility upfront by using a custom Handlebars template: https://github.com/webpro/release-it/blob/master/docs/recipes/auto-changelog.md.

webpro:
Closing this due to inactivity. Feel free to re-open if there's a need.

PR #394, opened 2018-11-13
ci: test Node.js 11

DanielRuf:
We should also test the current stable Node.js version which is currently 11.


webpro:
Thanks, @DanielRuf! I don't necessarily want to support non-LTS versions, but it definitely doesn't hurt to test release-it in the latest & greatest!

PR #396, opened 2018-11-19
Templates in commit and tag messages

PalatinCoder:
Make variables available in `src.commitMessage` and `src.tagAnnotation` by using the template helper, just as in the command hooks.

On Windows, this is somewhat unstable as cmd.exe can't handle the new lines (if any) in the actual `git tag` or `git commit` command. shell.js has an issue for that: shelljs/shelljs#175. I tried to adjust the readme to reflect this limitation. For now, you could only work around this by using a bash (either Git Bash or WSL).

The default configuration however still works fine, as this change also should be fully backwards compatible.

This PR is part of the feature discussed in #380 and #376

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,8 @@ Notes:
 
 - Only the settings to override need to be in `.release-it.json`. Everything else will fall back to the [default configuration](conf/release-it.json).
 - You can use `--config` if you want to use another path.
+- In `commitMessage` and `tagAnnotation` you can use variables (like template strings). These are the same variables as in the [command hooks](#command-hooks). 
+  However, be aware that these variables might inject new lines into the actual commands, and their execution will likely fail on windows.
 
 Any option can also be set on the command-line, and will have highest priority. Example:
 
--- a/lib/tasks.js
+++ b/lib/tasks.js
@@ -4,7 +4,7 @@ const { bump, runTemplateCommand: run, pushd, copy, npmPublish, popd, mkTmpDir }
 const git = require('./git');
 const githubClient = require('./github-client');
 const prompt = require('./prompt');
-const { truncateLines } = require('./util');
+const { truncateLines, template } = require('./util');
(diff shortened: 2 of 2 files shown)


webpro:
Thanks, @PalatinCoder! For this, I was thinking more of a result like this:

```
test('format', t => {
  t.equal(format('release v%s', { version: '1.0.0' }), 'release v1.0.0');
  t.equal(format('release v${version}', { version: '1.0.0' }), 'release v1.0.0');
  t.equal(format('release v${version} (${name})', { version: '1.0.0', name: 'foo' }), 'release v1.0.0 (foo)');
  t.end();
});
```

All calls to `format()` would need to be updated, resulting in `format` having the same interface as the `template` helper. Then they can be merged (if input contains `%` it calls `util.format`, otherwise `_.template`). I think the result is clean with only a single helper. How about it?

webpro:
@PalatinCoder, as I had some time for this I just went ahead and implemented my proposal in 261e9d0a109702a7627852e72b17ebedb4a2e7b4 as part of [v8.1.0](https://github.com/webpro/release-it/releases/tag/8.1.0). Hope you like it :-)

PalatinCoder:
@webpro Well, having only one method for formatting string is definitely a cleaner solution. But still, `version` is the only variable available for formatting in commit message or tag annotation, isn't it? My original intention with this PR was to make more variables (at least changelog) available. We would still need to pass them to the format helper now, wouldn't we?

webpro:
Oh, that's awkward! I basically removed some code, the config object as a whole should be available in the template now (in [v8.2.0](https://github.com/webpro/release-it/releases/tag/8.2.0)). Apologies for the confusion.

PalatinCoder:
Tested working!
Thank you so much @webpro! 👍 🎉

webpro:
Thanks @PalatinCoder, for testing so quickly and your response!

Issue #412, opened 2018-12-11
I can't use --preRelease=beta and conventional:angular together


dani8art:
When I try to release a beta from my package using `"increment": "conventional:angular"` it always releases a prepatch version. I think this is because when I use `--preRelease=beta`, `"increment": "prerelease"` is always set. So `release-it` skips version conventional calculation.

If i execute: 

```shell
release-it --dry-run --debug --preRelease=beta
```

then: 

```shell
release-it:config { _: [],
  release-it:config   'dry-run': true,
  release-it:config   d: true,
  release-it:config   dryRun: true,
  release-it:config   debug: true,
  release-it:config   e: true,
  release-it:config   preRelease: 'beta',
  release-it:config   increment: 'prerelease',
  release-it:config   preReleaseId: 'beta',
```
if I execute: 

```shell
release-it --dry-run --debug
```

then: 

```
release-it:config { _: [],
  release-it:config   'dry-run': true,
  release-it:config   d: true,
  release-it:config   dryRun: true,
  release-it:config   debug: true,
  release-it:config   e: true,
  release-it:config   increment: 'conventional:angular',
```

release-it version: v8.4.0
release-it config:
```
{
    "increment": "conventional:angular",
    "non-interactive": true,
    "requireUpstream": false,
    "preReleaseId": "canary",
    "requireCleanWorkingDir": false,
    "src": {
        "commitMessage": "ci: %s [ci skip]",
        "tagAnnotation": "%s"
    },
    "npm": {
        "publish": false
    },
    "github": {
        "release": true,
        "releaseName": "%s",
        "tokenRef": "GITHUB_TOKEN",
        "draft": false
    },
    "changelogCommand": "git log --pretty=format:'* %s (%h)' [REV_RANGE]"
}
```

How could I do that?

dani8art:
I would like to suggest: 

if we contact: `"increment": "conventional:angular:prerelease",`

then when we `const [system, preset, prerelease] = split(':')` we will have three var, and we can decide with:

```js
const recommendedType = await getConventionalRecommendedBump(preset);
if (prerelease) {
  const preReleaseMapper = { patch: 'prepatch', minor: 'preminor', major: 'premajor' };
  recommendedType = preReleaseMapper[recommendedType];
}
```

webpro:
[...]

PR #413, opened 2018-12-11
Modify increment calculation for supporting conventional and prerelease #412

dani8art:
related #412


webpro:
Thanks @dani8art! That seems to work fine. However, I don't think the code has become more readable. Coincidentally, I was working on a major release for which I've just released [v9.0.0-beta.0](https://github.com/webpro/release-it/releases/tag/9.0.0-beta.0). The version parser is refactored quite a bit, what do you think? It now boils down to this:

https://github.com/webpro/release-it/blob/master/lib/version.js#L41-L73

Please let me know whether this works for your use cases as well! Can we move on with this and squeeze the last bugs out?

dani8art:
It's so cool, this new release v9.0.0-beta.0 works fine for me.

Many thanks, @webpro.

webpro:
Thanks, @dani8art! I'm going to close this PR, and backport the new solution to v8 if I have some time today. Again, thanks for your PR: it was motivating and inspiring.

PR #420, opened 2018-12-16
Added additional file types, i.e. IDE, logs and local env files, to .…

nidkil:
Added additional files to .gitignore file, i.e. local env files, log files and IDE files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,25 @@
+.DS_Store
 .nyc_output
 .stage
 .tmp*
 coverage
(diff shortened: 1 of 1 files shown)


nidkil:
@webpro hmmmmm, I understand what you are saying, but I think a lot of people will not have things setup this way. If you look at project like Vue they have it setup this way to be as easy as possible for people to get started. I feel this, like the getting up & running instructions :-), help others contribute more easily.

Issue #417, opened 2018-12-14
Feature: add links to GitHub and NPM


nidkil:
Git adds the link to the GitHUb repo when a push has completed.

```
$ git push -u origin master
Enumerating objects: 112, done.
Counting objects: 100% (112/112), done.
Delta compression using up to 4 threads.
Compressing objects: 100% (95/95), done.
Writing objects: 100% (112/112), 555.16 KiB | 4.37 MiB/s, done.
Total 112 (delta 23), reused 0 (delta 0)
remote: Resolving deltas: 100% (23/23), done.
To https://github.com/nidkil/vue-test-plugin.git
 * [new branch]      master -> master
Branch 'master' set up to track remote branch 'master' from 'origin'.
```

In Webstorm and some consoles this link is clickable, which makes it very easy to go to the repo and do a quick check. It would be cool if release-it could add links to GitHub and NPM when the release process completes succsfully. Can I put in a pull request?

webpro:
I was thinking about exactly the same thing, so your PR is very welcome! Thanks in advance.

nidkil:
Okay I will have a look at it.

nidkil:
I have submitted 3 pull requests for this issue.

webpro:
In [v9.2.0](https://github.com/webpro/release-it/releases/tag/9.2.0).

PR #422, opened 2018-12-16
Show GitHub and npm urls

nidkil:
This is related to [issue #417](https://github.com/webpro/release-it/issues/417).

Added message that is displayed on completion with URLs to go to the GitHub repository and the NPM package. In a lot of terminals this URLs are displayed as clickable making it easy to jump to them and do a quick check.


nidkil:
I forgot to merge the upstream before making the changes. Fixed the code after merging upstream.

nidkil:
Fixed failing test.

lib/config.js
```diff
@@ -127,7 +127,8 @@ class Config {
       version,
       name,
       private: isPrivate,
-      publish: !!name
+      publish: !!name,
+      packageUrl: 'https://www.npmjs.com/package/' + name
```
webpro: Please add a helper to the `/lib/npm` module to provide this url.

lib/tasks.js
```diff
@@ -271,6 +271,19 @@ module.exports = async options => {
       await run(`!rm -rf ${dist.stageDir}`);
     }
 
+    if (github.release) {
+      getSpinner().stopAndPersist({
```
webpro: My bad, but please use the `log` helper for this.

lib/tasks.js
```diff
@@ -271,6 +271,19 @@ module.exports = async options => {
...
+        symbol: `❗`,
+        text: `To Github repository ${config.options.repo.remote}`
```
webpro: * I'd prefer a simple `🔗 [url]` here. I think it's clear from the url alone what we're linking to.
* The url should be coming from the return value of `release`: `releaseInfo.html_url` (not the repo url, which could be a non-web url, but the newly created release page).

lib/tasks.js
```diff
@@ -271,6 +271,19 @@ module.exports = async options => {
...
+      });
+    }
+    if (npm.publish) {
+      getSpinner().stopAndPersist({
+        symbol: `❗`,
+        text: `To NPM package ${config.options.npm.packageUrl}`
+      });
```
webpro: Something like this should be enough here:

```
log(`🔗 ${npm.getPackageUrl()}`);
```

lib/config.js
nidkil: Will do.

lib/tasks.js
nidkil: Okay.

lib/tasks.js
nidkil: Okay.

webpro:
Thanks @nidkil! The new feature is released in [v9.2.0](https://github.com/webpro/release-it/releases/tag/9.2.0).

PR #460, opened 2019-02-01
Add the `-u` flag to the `conventional-changelog` example

real34:
Otherwise the changelog would be empty


webpro:
Thanks @real34! I'll _release-it_ shortly.

PR #471, opened 2019-02-23
Support publishConfig.registry

n4bb12:
This PR adds support for npm's `publishConfig.registry` option.

Related docs:
- https://docs.npmjs.com/files/package.json#publishconfig
- https://docs.npmjs.com/misc/config#registry

Currently `release-it` fails when publishing to a registry other than the default because [this sanity check](https://github.com/webpro/release-it/blob/10.1.0/lib/npm.js#L58) does not take take into account `publishConfig.registry`.

Additionally, the [output after a successful publish](https://github.com/webpro/release-it/blob/master/lib/tasks.js#L164) makes [strong assumptions on the default registry](https://github.com/webpro/release-it/blob/10.1.0/lib/npm.js#L67).

### DIFF
--- a/lib/npm.js
+++ b/lib/npm.js
@@ -1,12 +1,13 @@
 const semver = require('semver');
 const _ = require('lodash');
+const urlJoin = require('url-join');
 const { rejectAfter } = require('./util');
 const debug = require('debug')('release-it:npm');
 const { npmTimeoutError, npmAuthError } = require('./errors');
 
 const REGISTRY_TIMEOUT = 10000;
 const DEFAULT_TAG = 'latest';
-const NPM_BASE_URL = 'https://www.npmjs.com/package/';
+const NPM_DEFAULT_REGISTRY = 'https://www.npmjs.com';
 
 const noop = Promise.resolve();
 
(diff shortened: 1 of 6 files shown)


lib/npm.js
```diff
@@ -55,16 +56,21 @@ class npm {
   }
 
   isAuthenticated() {
-    return this.shell.run('npm whoami').then(() => true, () => false);
+    const registry = this.getRegistry(this.options);
+    return this.shell.run(`npm whoami --registry ${registry}`).then(() => true, () => false);
```
n4bb12: At this point we could omit the flag if it's the default registry. I'm always appending it for simpler code, but ofc something we could change if you want that.

webpro:
Hi @n4bb12, great PR! Very much appreciated. The one small thing that would make this PR 💯 is a test reflecting that `publishconfig.registry` is read from `package.json` by default. Could you please add a test for this, perhaps in one of the more "integration" tests in `test/tasks.js`?

n4bb12:
Is that what you had in mind?

webpro:
Yes, thanks @n4bb12! I'll _release-it_ shortly.

webpro:
Released in [v10.2.0](https://github.com/webpro/release-it/releases/tag/10.2.0)

Issue #472, opened 2019-02-24
scripts.changelog does not actually executed in --dry-run mode


1natsu172:
Thanks for the best release tool.

Version: `release-it@10.1.0`

part of `.release-it.json`

```json
  "scripts": {
    "changelog": "conventional-changelog -p angular | tail -n +3",
    "beforeStage": "conventional-changelog -p angular -i CHANGELOG.md -s",
    "afterRelease": "echo -e \"Successfully released ${name} v${version} to ${repo.repository}.\n\nNext: Let's write release notes on GitHub :)\"; open https://github.com/1natsu172/Outside-YouTube-Player-Bar/releases"
  },
```

---

I want to see the content of CHANGELOG before release, so I run with `--dry-run`, but it is displayed  `Empty changelog`.(Actually not empty. After production release, CHANGELOG was output properly…)

![](https://i.gyazo.com/e2090e54cb77c86894fb12b314d65e82.png)

I can understand that the `beforeStage` and `afterRelease` commands will not be executed because it can not handle on "release-it" side… 🤔 

Is it intentional that `scripts.changelog` is not executed? (or is it wrong to check CHANGELOG by dry-run?)

webpro:
Hi @1natsu172! It intentionally does not execute any custom user scripts from `scripts.*` as they *may* write things, although `scripts.changelog` *should* only read.

So yeah, I agree a changelog should be generated in a dry run as well. You up for a PR with a fix? :-)

1natsu172:
Thanks for fast reply!

First of all I wanted to confirm to you whether this behavior was intentional.

> You up for a PR with a fix? :-)

I think will submit fix PR tomorrow, if can do it well 😃

webpro:
Released in [v10.2.0](https://github.com/webpro/release-it/releases/tag/10.2.0)

PR #473, opened 2019-02-25
change to execute scripts.changelog even in dry-run mode

1natsu172:
Hi @webpro , this PR about #472 📝 

Contents of tweak: The caller side can specify `{ isWrite: false }` as option in `shell.runTemplateCommand`.

When using "Conventional-Changelog" for the `scripts.changelog`, the content of this PR(#460) is actually necessary. This is probably because the "release-it" does not actually execute the git operation in dry-run mode, so the "Conventional-Changelog" side can't detect the changes. 💭 

About `-u` flag is [here](https://github.com/conventional-changelog/conventional-changelog/blob/97ad96fe893bd5b40ec52e24fe46f4f9cd357a1a/packages/conventional-changelog-cli/cli.js#L40)

### DIFF
--- a/lib/changelog.js
+++ b/lib/changelog.js
@@ -18,7 +18,8 @@ class Changelog {
           }
         });
       } else {
-        return this.shell.runTemplateCommand(command);
+        const option = { isWrite: false };
+        return this.shell.runTemplateCommand(command, option);
       }
     }
     return noop;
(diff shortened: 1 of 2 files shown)


webpro:
Perfect, thanks @1natsu172! I'll _release-it_ shortly.

PR #478, opened 2019-03-01
add dotenv for local .env file variables

spentacular:
Hey @webpro thanks for the great tool!

This is just a tiny tweak to add [`dotenv`](https://github.com/motdotla/dotenv) and run it in `release.js` to allow `process.env` to pick up a local `.env` file variables.

This should help if someone has both a GitHub Enterprise and personal GitHub project both running release-it, requiring two different keys, but it's also a habit I've picked up since it's incredibly common for [The Twelve-Factor App](http://12factor.net/config).

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,6 +249,12 @@ The "Releases" tab on GitHub projects links to a page to store the changelog. To
 export GITHUB_TOKEN="f941e0..."
 ```
 
+or inside an `.env` file (make sure to add this file to your `.gitignore`):
+
+```
+GITHUB_TOKEN="f941e0..."
+```
+
 Do not put the actual token in the release-it configuration. It will be read from the `GITHUB_TOKEN` environment
 variable. You can change this variable name by setting the `github.tokenRef` option to something else.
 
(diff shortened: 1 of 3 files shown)


webpro:
Hi @spentacular, thanks. I'm not sure whether it should be release-it's responsibility to read the `.env` file. There are multiple ways to achieve this, one is to add e.g. `-r dotenv/config` to the (npm) script that executes release-it. Would something like that work for you as well?

spentacular:
Hey @webpro, I do agree there are other ways, but I'll give my pitch really quick 😀

I'm personally not a fan of global things and try to keep those clean. `.env` helps isolate things to their project where they are used at. Typically you also commit a `.env.example` or similar file with only the keys as well, which helps keep a good source of truth for what variables need to exist if someone were to start new or move CI servers. [random example file](https://github.com/mrgodhani/slack-bot/blob/master/.env.example). Even if there isn't an `.env.example` file, it's also much easier to look at your own repo if you're onboarding a new coworker and remember "Oops, I forgot about the environment variables!".

My personal backup plan is to just use `direnv` to manage keeping those variables next to the code or do the `-r dotenv/config` for our use cases. I'm just biased towards ways that allow variables to be set without adding them to my shell profile/config.

webpro:
I'm not disagreeing with those principles :-)

spentacular:
Sounds good! Thanks again for the great tool.

PR #482, opened 2019-03-04
Improve docs on auto-changelog usage

### LINKED ISSUE #480: Improve explanation in README on how to use auto-changelog as a companion
Hello, I am testing out `release-it`, considering using it for one of my next projects. Seems promising so far, but I've encountered a weird behavior (at least to me) while following the steps suggested in the readme:

> Additionally, `scripts.beforeStage` will update the `CHANGELOG.md` with each release to get included with the release commit. Obviously this can be omitted if the project does not keep a `CHANGELOG.md` or similar.

What I find odd is: since the `CHANGELOG.md` is being updated *before* the version bump, `auto-changelog` won't know about the upcoming new version, and things will go under the `Unreleased` bullet. What should happen, IMO, is the changelog should reflect the new version at the same time the version is being released. If somehow I use another script hook (other than `beforeStage`) to do this, then the changelog will probably be correct but now it will only be generated after the release is done, effectively making the release itself "changelog-less".

I am honestly lost here, so I would be super happy with an explanation and I think README should be improved so that others don't become confused as I currently am.

On a side note, I think the script hooks deserve a better documentation. By their names, I would think that `afterBump` runs after the `package.json` version is bumped and committed, but all scripts until `beforeStage` run first and then `release-it` asks me if I want to commit.

papb:
Fixes #480.

The main source of confusion was the fact that the `-p` option was missing in the auto-changelog usage suggestion.

I also added some explanations to the script hooks which would have helped me, so probably will help others.

By the way, what is the difference between the `afterBump` and `beforeStage` hooks? That is still not so clear to me. Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -334,7 +334,7 @@ A tool like [auto-changelog](https://github.com/CookPete/auto-changelog) is a gr
 {
   "scripts": {
     "changelog": "npx auto-changelog --stdout --commit-limit false -u --template ./changelog.hbs",
-    "beforeStage": "npx auto-changelog"
+    "beforeStage": "npx auto-changelog -p"
   }
 }
 ```
--- a/docs/recipes/auto-changelog.md
+++ b/docs/recipes/auto-changelog.md
@@ -14,7 +14,7 @@ Example configuration in the release-it config:
 
 ```json
 "scripts": {
-  "beforeStage": "npx auto-changelog",
+  "beforeStage": "npx auto-changelog -p",
   "changelog": "npx auto-changelog --stdout --commit-limit false --unreleased --template ./preview.hbs"
 }
 ```


webpro:
* The `-p` should not be there I think.
* The `afterBump` and `beforeStage` are basically legacy from earlier on. The next major (v11) is a nice opportunity to remove one of them.

papb:
@webpro - I respectfully disagree - the `-p` option should be there.

From the [auto-changelog] README:

> Using `-p` or `--package` uses the `version` from `package.json` as the latest release, so that all commits between the previous release and now become part of that release. Essentially anything that would normally be parsed as `Unreleased` will now come under the `version` from `package.json`

This means that without `-p`, everything that is being released will be generated as *Unreleased* which is incorrect. And I've seen it happen, as I said in my issue #480, this was a source of confusion to me.

Can you please reconsider merging this PR? It's just to prevent confusion for other newcomers to this project. Thanks :)

webpro:
My sincere apologies @papb, you are entirely correct. Not sure what my settings where at the time, but I also did https://github.com/release-it/release-it/blob/master/docs/recipes/auto-changelog.md then. If you could add `-p` in that document as well, and remove the script hooks copy changes (the prose does not really align with the rest of the docs), I'm happy to merge this and credit you for finding this!

papb:
@webpro - Done and thank you :)

webpro:
Thanks @papb!

PR #489, opened 2019-03-06
Add note in readme about JS config files

papb:
Using `.release-it.js` instead of `.release-it.json` was something that helped me immensely, I think this possibility deserves to be documented directly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,7 @@ Notes:
 - Only the settings to override need to be in `.release-it.json` (or `package.json`). Everything else will fall back to
   the [default configuration](conf/release-it.json).
 - You can use `--config` if you want to use another path for `.release-it.json`.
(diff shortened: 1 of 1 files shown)


webpro:
A file `.release-it.js` is not used by release-it, it has no effect.

papb:
@webpro I see now that I wasn't very clear in the description of this PR, sorry about that. If you take a look at the proposed changes, I think you will understand.

Basically, since `release-it` allows passing a custom path to the configuration file, and since it is just `require()`-ed by `release-it` internally, this means that using a standard `.js` file also works. In other words, the JSON format is not mandatory for the config file, a standard `.js` file also works. The goal of this PR is to clarify this fact. This was very useful to me, for example.

webpro:
Oh I did not even realize you could do that, ha! Thanks for pointing it out (again).

PR #507, opened 2019-04-20
Move tests which require GITHUB_TOKEN to IT tests

seanodell:
These tests seem more like integration tests that not everyone should run. I did add a call to them as the first step in the release step, but not sure where else they should be run. This change will allow `npm test` to pass for anyone who is just cloning and running `npm install`.


webpro:
All tests in `tasks.js` are integration tests. I agree this could be made clear somehow. On the other hand, there are many integration tests all around, such as the Git tests which are doing disk IO. So that's why I didn't really feel a need to distinguish between purely test vs integration tests, but keep them in a single file per module.

Of course all tests should run without problems. In Travis CI (https://travis-ci.org/release-it/release-it) I added dummy values for `GITHUB_TOKEN` etc. Much like the solution I implemented (in the v11 branch) just before you opened this PR: https://github.com/release-it/release-it/blob/v11/test/util/setup.js.

So I'm going to close this PR, but I do like to discuss this topic after v11 is out.

seanodell:
Okay, makes sense. I don't think that was intended to be a true integration test if dummy values are preferred. I like your approach better.

webpro:
It's indeed not truly "end to end" or something. No network traffic is happening. There is disk read/writes, though (e.g. Git is not mocked).

PR #508, opened 2019-04-20
Support package-lock

seanodell:
It's possible to include package-lock.json in .release-it.json in pkgFiles, but since release-it depends on node engine >= 8 and package-lock.json has been default since 5 (and should be checked into source control), I think it be expected behavior for a lot of people. Possibly people don't realize that their package-lock.json is not getting updated, so this could be a somewhat hidden issue for a lot of people.

Take with a grain of salt. I'm by no means an expert at node or npm. Let me know how I could handle this better for you.

### DIFF
--- a/conf/release-it.json
+++ b/conf/release-it.json
@@ -1,6 +1,6 @@
 {
   "preReleaseId": null,
-  "pkgFiles": ["package.json"],
+  "pkgFiles": ["package.json", "package-lock.json"],
   "use": "git.tag",
   "scripts": {
     "beforeStart": null,
--- a/test/tasks.js
+++ b/test/tasks.js
@@ -78,13 +78,18 @@ test.serial('should run tasks without throwing errors', async t => {
 
 test.serial('should run tasks with minimal config and without any warnings/errors', async t => {
   gitAdd('{"name":"my-package","version":"1.2.3"}', 'package.json', 'Add package.json');
+  gitAdd('{"name":"my-package","version":"1.2.3"}', 'package-lock.json', 'Add package-lock.json');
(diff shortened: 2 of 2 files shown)


webpro:
Thanks @seanodell, agreed. Could you please update the tests so it asserts `package-lock.json` is correctly updated only once? These tests are already quite comprehensive, so the less noise the better.

By the way, around next week v11 will be released, which delegates bumping these files to `npm version`.

seanodell:
> Could you please update the tests so it asserts `package-lock.json` is correctly updated only once?

Sure, I'll take a look. Not exactly sure of what to test for, but my guess is that as I updated tests that operated on package.json, I missed one?

webpro:
Sorry, I wasn't clear. I meant we can assert that `package-lock` is correctly updated in only a single test, just once. We don't need to make the same assertion three or more times.

seanodell:
Done. Had to keep one small change in another test, but the primary assertion happens in 'should run tasks with minimal config and without any warnings/errors'.

webpro:
Thanks @seanodell. Just tried this in a repo with `package-lock.json`:

```
WARNING Could not bump package-lock.json
✔ 
WARNING Could not stage package.json package-lock.json
```

This situation is also quite common, so I don't really dig this change. It may surprise people, even though it's not a breaking change (and obviously I should update my `pkgFiles` array). If we can wait just a few more days the problem will be auto-resolved in v11: https://github.com/release-it/release-it/blob/5830ba59430f73907627175b9bd3b965da6c5ea3/lib/plugin/npm/npm.js#L67

Issue #495, opened 2019-03-18
Github Release publish only after asset upload


Luehrsen:
https://github.com/release-it/release-it/blob/83a40f6574b4367eddfeb841e69350efc2f56d52/lib/github.js#L96

Hey, amazing work on release-it! 

We are currently working with github actions and the release trigger. If I understand the code correctly, release-it handles Github releases like this: 
1) Create the release
2) Publish the release (if not draft)
3) Add assets to the release

This is unfortunate, as this omits release assets to be present in the [Release Even Payload](https://developer.github.com/v3/activity/events/types/#releaseevent) for GitHub Actions. 

So a more convenient workflow would be: 
1) Create the release as draft
2) If exist, add assets to the release
3) If applicable, publish the release

webpro:
I see the issue. Feel free to open a PR to make it happen.

Luehrsen:
Hey @webpro, 

this issue has somehow sneaked back into production rendering all github `on:release` actions useless, that rely on assets being present. 

To shortly reiterate the issue: 

1. release-it creates the github release [here](https://github.com/release-it/release-it/blob/496ce1c5b7c29d8bf68867c9afe30ccec42f8236/lib/plugin/github/GitHub.js#L110)
2. The release is immediately published, prompting the github actions to trigger, the assets cannot be found
3. release-it then uploads the assets.

This was broken in https://github.com/release-it/release-it/commit/d516bdbc2d1c9466fe0e044494d258ac9bcd6246#diff-5522b930e6cbd1617754a2cb7b188ee2

webpro:
Unfortunately the refactoring you are referring to has basically reverted to the behavior that's problematic for you. My apologies for introducing a breaking change in a minor update!

See https://github.com/release-it/release-it/pull/659 to read what started this change.

However, I think the new behavior supports your use case as well. The idea is to draft a release including the assets (this should not trigger the action):

```
release-it --github.release --github.draft --github.assets=file
```

And then to publish the release and trigger the action, for example:

```
release-it --github.release --no-npm.publish --no-increment --no-git
```

Unfortunately the GitHub API doesn't support the creation of a release with the assets included in one go, and since the refactoring release-it is more flexible with regards to updating the latest release/assets. Unfortunately for your use case this means you would need two separate commands now.

Let me know if this works for you.

Regardless, things should obviously be documented.

PR #512, opened 2019-04-26
Add initial draft step to GitHub release

Luehrsen:
This PR moves the actual publishing of the GitHub release to after the release assets have been uploaded.

This is important because Github Action triggers a release on publish, so it is possible that release assets might not be visible to the action, as they are yet to be uploaded.

see #https://github.com/release-it/release-it/issues/495

### DIFF
--- a/lib/github.js
+++ b/lib/github.js
@@ -163,9 +163,10 @@ class GitHub extends Release {
     return `https://${host}/${repository}/releases/tag/${tag}`;
   }
 
-  maybePublishRelease() {
-    const { draft } = this.options;
+  maybePublishRelease({ version } = {}) {
+    const { draft, tagName } = this.options;
     const { owner, project: repo } = this.repo;
+    const tag_name = format(tagName, { version });
 
     if (draft) {
       // The user actually wanted a draft so we don't publish.
@@ -178,12 +179,14 @@ class GitHub extends Release {
           owner,
           repo,
           draft,
-          release_id: this.releaseId
+          release_id: this.releaseId,
+          tag_name
         };
         debug(options);
         const response = await this.client.repos.updateRelease(options);
         debug(response.data);
         this.log.verbose(`octokit releases#publishRelease: done (${response.headers.location})`);
+        this.releaseUrl = response.data.html_url;
         return response.data;
       } catch (err) {
         return this.handleError(err, bail);
--- a/lib/github.js
+++ b/lib/github.js
@@ -67,7 +67,7 @@ class GitHub extends Release {
     return ghClient;
   }
 
-  async release({ version, isPreRelease, changelog } = {}) {
+  async draftRelease({ version, isPreRelease, changelog } = {}) {
     const { tagName, releaseName, releaseNotes, isDryRun } = this.options;
     const tag_name = format(tagName, { version });
     const name = format(releaseName, { version });
--- a/lib/github.js
+++ b/lib/github.js
@@ -81,7 +81,6 @@ class GitHub extends Release {
       return noop;
     }
 
-    const { draft } = this.options;
     const { owner, project: repo } = this.repo;
 
     return this.retry(async bail => {
@@ -93,14 +92,15 @@ class GitHub extends Release {
           name,
           body,
           prerelease: isPreRelease,
-          draft
+          draft: true
         };
         debug(options);
         const response = await this.client.repos.createRelease(options);
         debug(response.data);
-        const { html_url, upload_url } = response.data;
+        const { html_url, upload_url, id } = response.data;
         this.log.verbose(`octokit releases#createRelease: done (${response.headers.location})`);
         this.releaseUrl = html_url;
+        this.releaseId = id;
         this.uploadUrl = upload_url;
         this.isReleased = true;
         return response.data;
@@ -161,6 +161,36 @@ class GitHub extends Release {
     const tag = format(this.options.tagName, { version });
     const { host, repository } = this.repo;
     return `https://${host}/${repository}/releases/tag/${tag}`;
+  }
+
+  maybePublishRelease() {
+    const { draft } = this.options;
+    const { owner, project: repo } = this.repo;
+
+    if (draft) {
+      // The user actually wanted a draft so we don't publish.
+      return;
+    }
+
+    return this.retry(async bail => {
+      try {
+        const options = {
+          owner,
+          repo,
+          draft,
+          release_id: this.releaseId
+        };
+        debug(options);
+        const response = await this.client.repos.updateRelease(options);
+        debug(response.data);
+        const { html_url } = response.data;
+        this.log.verbose(`octokit releases#publishRelease: done (${response.headers.location})`);
+        this.releaseUrl = html_url;
+        return response.data;
+      } catch (err) {
+        return this.handleError(err, bail);
+      }
+    });
   }
 }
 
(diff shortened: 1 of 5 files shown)


webpro:
Thanks, @Luehrsen! Looks good. Could you please add coverage for this in the tests?

Luehrsen:
Will do...

Luehrsen:
As far as I can see 'github should release and upload assets' https://github.com/release-it/release-it/blob/master/test/github.js#L35 already covers this code. 

The fail-case would be, that a release is drafted although draft is not `true`. The tests already test if a tag_name is assigned. The tag_name itself is only assigned on published releases, as drafts do not yet have a valid tag. 

As for some of the other tests that currently fail: I have no idea how to fix them. AFAIK the also fail on master.

webpro:
* You can see what parts of the code aren't covered yet with `npm run coverage` (results in the `coverage` dir).
* The tests should run without problems: https://travis-ci.org/release-it/release-it/builds
* I can see the failing tests from this branch here: https://travis-ci.org/release-it/release-it/jobs/524915444#L634. The first error makes sense: the `draft` prop is now `true` (not `false` anymore) after your changes. And so on. Or are you referring to other failing tests?

Luehrsen:
Ah, this is my first time working with ava (and TDD is not part of my usual workload), so sorry for the confusion. 

I have started to implement the tests, but I've hit somewhat of a roadblock: updateRelease on L184 returns a rejected promise, that leads the following code to fail. Any idea on how to get that to work? 

Thanks!

Luehrsen:
Found my issue: I need a stub. Working on it now...

Luehrsen:
I still have quite some issues while debugging the tests. I'm not seeing the same fails locally as travis is seeing. For me, all git related tests fail.

```
  should run tasks without throwing errors

  /Projekte/wp_development/projects/release-it/tmp/9953082f-daad-4cd0-b2f1-3496b96ad725/lib/shell.js:44

  Rejected promise returned by test. Reason:

  Error {
    message: `No refs in common and none specified; doing nothing.␊
    Perhaps you should specify a branch such as 'master'.␊

    fatal: the remote end hung up unexpectedly␊
    error: failed to push some refs to '/Projekte/wp_development/projects/release-it/tmp/fe5eda00-09dc-4e9b-b04f-a7e280d3fc70'␊
    `,
  }

```

I believe there is some issue in the git config on my system, but I can't figure out what. (And I haven't had any issues up until now...)

Luehrsen:
Found the issue: My local git config defaulted to 'matching' as push default, but this setup expects 'simple'. 

See: https://git-scm.com/docs/git-config#Documentation/git-config.txt-pushdefault

test/github.js
```diff
@@ -58,12 +58,23 @@ test('should release and upload assets', async t => {
...
 
+  t.is(releaseResult.tag_name, 'v' + version);
+  t.is(releaseResult.name, 'Release ' + version);
+
   t.is(github.isReleased, true);
   t.is(github.getReleaseUrl(), 'https://github.com/webpro/release-it-test/releases/tag/v2.0.1');
 
   t.is(releaseResult.tag_name, 'v' + version);
   t.is(releaseResult.name, 'Release ' + version);
-  t.is(githubRequestStub.callCount, 1);
```
webpro: We can keep this assertion.

lib/tasks.js
```diff
@@ -130,8 +130,12 @@ const runTasks = async (opts, injected = {}) => {
       github.release && github.releaseNotes && logPreview(log, 'release notes', await ghClient.getNotes());
       const ghRelease = () => ghClient.release({ version, isPreRelease, changelog });
       const ghUploadAssets = () => ghClient.uploadAssets();
+      const ghPublishRelease = () => ghClient.maybePublishRelease();
       if (isInteractive) {
-        const release = () => ghRelease().then(() => ghUploadAssets());
+        const release = () =>
+          ghRelease()
+            .then(() => ghUploadAssets())
+            .then(() => ghPublishRelease());
```
webpro: Please also update the flow for `isInteractive === false` below. This will then fix the `draft: true` assertions in `test/tasks.js` ;-)

The flows are slightly different. After prompted to release to GitHub, there should be a single `task` to complete. But in non-interactive mode, it's nice to separate the actions to better indicate the progress.

lib/tasks.js
```diff
@@ -130,8 +130,12 @@ const runTasks = async (opts, injected = {}) => {
       github.release && github.releaseNotes && logPreview(log, 'release notes', await ghClient.getNotes());
       const ghRelease = () => ghClient.release({ version, isPreRelease, changelog });
       const ghUploadAssets = () => ghClient.uploadAssets();
+      const ghPublishRelease = () => ghClient.maybePublishRelease();
```
webpro: Maybe with `maybePublishRelease` it makes sense to rename `release` to `draftRelease`?

Luehrsen:
Hey @webpro, I hope you had a great weekend. 

I've renamed the release function and added the non-interactive mode. But now I've got some failed tests which I believe have to do with the order of calls being made. Maybe even a race condition? I really don't understand what is going on there, maybe you can have a look.

webpro:
The situation is that we are now going from two to three calls to GitHub (or the stub in test runs). The bug in the PR is that `maybePublishRelease` overwrites the url (`this.releaseUrl = html_url`) from `draftRelease` while it's `undefined`. So we can just remove this line from the `draftRelease` as it was already set.

Please find attached the patch file. Remove the `.txt` extension and apply with `patch -p0 -i pr512.patch`:

[pr512.patch.txt](https://github.com/release-it/release-it/files/3126402/pr512.patch.txt)

I've also added a few assertions for both related calls. Hopefully this clears things up a little. If not, please do not hesitate to let me know :)

Luehrsen:
The tests are all green, but I've actually found an issue: 

The release_url that is getting printed to the command line is wrong. It is always the draft URL, no matter if it is released or not. So we actually need to overwrite the releaseUrl. (Or find another method to overwrite that.)

webpro:
Are the url's different? What is printed? It should be `https://github.com/${owner}/${repoName}/releases/tag/{tag}`

Luehrsen:
The format is correct, the tag is wrong. 

This is what gets printed: 
`🔗 https://github.com/luehrsenheinrich/release-it-test/releases/tag/untagged-1e51fa1f91e3c1b00d61`

This is what it should be:
`https://github.com/luehrsenheinrich/release-it-test/releases/tag/1.0.12`

webpro:
Ah I see. Then we need to inspect the response from the `repos.updateRelease` call and hopefully the url is in there...?

You can use `DEBUG=release-it:* release-it` to have this response logged.

lib/github.js
```diff
@@ -183,9 +183,7 @@ class GitHub extends Release {
         debug(options);
         const response = await this.client.repos.updateRelease(options);
         debug(response.data);
-        const { html_url } = response.data;
```
Luehrsen: This is where the updated release URL would come out, this is why this override exists...

Luehrsen: *existed. This got removed by the patch. To we need that, but we also need that to play nice with the stubs. 🤔

webpro: OK now I get it. We can send the `tag_name` again with the `options` in the `repos.updateRelease` (so it will also come back with the stub response) and restore the line I removed here. Apologies for the confusion!

Luehrsen: Well, this should be it...

webpro:
Thanks a lot, great work! I've squashed your work into [v10.4.5](https://github.com/release-it/release-it/releases/tag/10.4.5).

Issue #518, opened 2019-05-04
Add `quiet` option to suppress output


danderson00:
I intend to have a Release 'em! option to do a dry run first to do a simple validation before doing an actual run and it would be great to suppress this output and only display any errors. This should stop me from being an idiot and forgetting to commit changes to one repo, causing it to stop after potentially already publishing some. 

Should be a fairly simple addition to the logger class and global options, I'll put together a PR.

webpro:
Just my initial feedback: it sounds a bit odd to suppress all output for a CLI tool like release-it. Is that what you want to open a PR for? Is there no other way from the outside to catch its output, or override the Logger?

danderson00:
The idea is that it will do a validation run quietly first, just to catch silly mistakes, then do the actual run. No point in displaying the output twice. I dunno, `--quiet` seems a reasonably common option for CLI tools?

I can't see a way to override the Logger class, and I guess I could shadow the console.log method, but that's a bit nasty. I have a very simple implementation with tests ready, I'll put a PR together now for your thoughts.

webpro:
Indeed, `--quiet` is extremely common in CLI tools. Maybe it's just me, but so many things can happen in a single run, let alone with custom user scripts, without seeing what's going on (fully), that I'm a bit hesitant towards this. I still do see use cases for this feature, just need to be careful with (presenting) it.

So your PR is very useful, will look into it. However, what I totally forgot about initially: release-it is built completely with dependency injection. One of the reasons is easier testing/mocking, but this is useful to "bring your own logger" as well. Here is the relevant code:

https://github.com/release-it/release-it/blob/dc9aa89106c008a488677e369163e7c2ffd14098/lib/tasks.js#L13-L24

So you can do:

```js
release(taskConfig.config, {
  log: new MyLogger()
})
```

This gives you even more control. Just make sure to implement the same methods.

danderson00:
Gotcha, that's not a bad solution. `index.js` doesn't currently pass on the `di` parameter, so I'd have to require `tasks` directly, but that's OK for now. I'll also have to do something with the spinner (`ora`), but it's also set from `di`, so that's possible.

I'm happy for this and the PR to be closed if you see no benefit. Thanks for the prompt response!

webpro:
Or even `sinon.createStubInstance(Log)` if you're lazy. Please leave the PR open, might still be useful sooner or later.

PR #519, opened 2019-05-04
Add quiet option

danderson00:
It's a relatively simple change. Up to you, I can find other workarounds if you're not keen on the change.

### DIFF
--- a/bin/release-it.js
+++ b/bin/release-it.js
@@ -11,17 +11,19 @@ const aliases = {
   h: 'help',
   i: 'increment',
   n: 'non-interactive',
+  q: 'quiet',
   v: 'version',
   V: 'verbose'
 };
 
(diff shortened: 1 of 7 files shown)


danderson00:
Resolves #518

webpro:
The PR is fine, not a big impact on the code overall.

However, I think simply overriding interactive mode by using `--quiet` is unexpected behavior (especially since interactive mode is the default). So I'd say (at least for a first release) it's better to explicitly have to provide both `-n` and `--quiet` arguments to run quietly. What do you think?

danderson00:
Leaving open for now, as per #518. If this is merged, https://github.com/release-it/release-it/blob/master/lib/spinner.js also needs to be updated to suppress output.

PR #523, opened 2019-05-08
npm.ping option to disable npm ping in case your repo doesn't support it

mafopen:
e.g., https://issues.sonatype.org/browse/NEXUS-13434

I prefer this option rather than extending the check for which errors to ignore. But let me know if you would like me to extend the regex instead?

### DIFF
--- a/conf/release-it.json
+++ b/conf/release-it.json
@@ -23,6 +23,7 @@
     "pushRepo": "origin"
   },
   "npm": {
+    "ping": true,
     "publish": true,
     "publishPath": ".",
     "access": null,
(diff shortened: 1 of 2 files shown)


webpro:
Thanks for the report. I'm hesitant towards adding configuration options. In this case, it may not be clear for a user why there's a `npm.ping` option at all. E.g. setting it to `false` if maybe the internet connection is down has undesired effects. Also, if the implementation of the prerequisite checks changes, this option may become irrelevant/incorrect.

Would something like this work here?

```
/(Error |code E)40[04]/
```

mafopen:
Fair enough, see https://github.com/release-it/release-it/pull/524

PR #529, opened 2019-05-22
Add a prepareRepo step

pierreroth64:
I would like to use `release-it` for my react-native app repository (android).
The expected workflow for my use case is: 
```
[1] bump version => [2] commit, tag and push tag => [3] build the apk => [4] release (and push the previously built apk) to github releases.
```
Currently there's no hook to build the apk between setting the tag (step [2]) and the actual gh release (step [4]).

This PR proposes a new step called `prepareRepo` (step [3]) and therefore an `afterPrepareRepo` hook. People that need to "build something" from the created tag (release artifacts such as .apk or any binary) may use this hook to call their own build script.

### DIFF
--- a/README.md
+++ b/README.md
@@ -551,6 +551,9 @@ name
 repo.remote, repo.protocol, repo.host, repo.owner, repo.repository, repo.project
 ```
 
+The `version` variable is available from `beforeBump` (it's only available in `beforeStart` if there is no need to
+prompt the user for the next version).
+
 ## Plugins
 
 Since v11, release-it can be extended in many, many ways. Please head over to [plugins](docs/plugins/README.md) for more
--- a/docs/plugins/README.md
+++ b/docs/plugins/README.md
@@ -4,6 +4,15 @@ Since v11, release-it has evolved into a pluggable task runner. It was previousl
 GitHub/GitLab releases, and npm packages, but this is no longer the case. As long as can either be written in Node.js,
 or executed from the shell, it can be integrated in the release-it process.
 
+## Contents
+
+- [Overview](#overview)
+- [Using a plugin](#using-a-plugin)
+- [Creating a plugin](#creating-a-plugin)
+- [Available & example plugins](#available--example-plugins)
+
+### Overview
+
 Plugins allow additional and custom actions in the release process, such as:
 
 - Publish the package to any registry (this is language-agnostic, e.g. Ruby, Python, ...).
@@ -48,7 +57,7 @@ This example uses the `release-it-bar` module and is configured in `package.json
     },
     "plugins": {
       "release-it-bar": {
-        "option": "value"
+        "key": "value"
       }
     }
   }
@@ -61,7 +70,7 @@ Alternatively, here's a `foo` plugin as a local module:
 {
   "plugins": {
     "./scripts/foo.js": {
-      "option": "value"
+      "key": "value"
     }
   }
 }
@@ -71,36 +80,17 @@ Alternatively, here's a `foo` plugin as a local module:
 
 To create a plugin, extend the `Plugin` class, and implement one or more release-cycle methods. See the "interface"
 below (where none of the methods is required). Any of these methods can be `async` (except for
-`getIncrementedVersionSync`). If you're interested in writing a plugin, please take a look at
+`getIncrementedVersionCI`). If you're interested in writing a plugin, please take a look at
 [the `runTasks` test helper](https://github.com/release-it/release-it/blob/master/test/util/index.js#L33-L54), to see
 how a plugin is integrated in the release process. Also see the
 [base `Plugin` class](https://github.com/release-it/release-it/blob/master/lib/plugin/Plugin.js) where the plugin should
 be extended from.
 
-## Interface
-
-```javascript
-class Plugin {
-  static isEnabled() {}
-  static disablePlugin() {}
-  init() {}
-  getName() {}
-  getLatestVersion() {}
-  getIncrementedVersionSync() {}
-  getIncrementedVersion() {}
-  beforeBump() {}
-  bump() {}
-  beforeRelease() {}
-  release() {}
-  afterRelease();
-}
-```
-
 ### Example
 
 This minimal example reads the current version from a `VERSION` file, and bumps it once the new version is known.
 
-```javascript
+```js
 class MyPlugin extends Plugin {
   getLatestVersion() {
     return fs.readFileSync('./VERSION').trim();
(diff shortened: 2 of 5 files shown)


webpro:
Thanks @pierreroth64! While your use case is a perfectly valid one, I think we should be careful with adding another method to already the broad matrix of plugins/life-cycle, and script hooks.

It seems what you need here is a hook in-between the `release` method of the core `git` and `github` plugins, right? And you prefer to use a script hook (i.e. not developing a plugin)?

Just to be sure, does building the APK require a tag? If it only requires the new `version` you could perhaps use the `beforeStage` script hook (and provide the `{version}` in the command).

pierreroth64:
Thanks for your quick feedback @webpro ! Yes I need to build **after** the version has been bumped *and* git-commited (core git plugin) but **before** release has been pushed to GH (core github plugin)

- What I understand from https://github.com/release-it/release-it/blob/master/lib/plugin/factory.js#L67, is that we cannot "register" a plugin in between core plugins, so unless I missed something, this is not possible with the current code as my `Build plugin` would implement the `release()` method but would require to be called within the core plugins sequence.
- As I need the version to be commited before calling my build script, the current hooks won't help me either.
That's why, my first proposal was to add a new step to "prepare repository" (`prepareRepo()` method) before actually "releasing" (`release()` method)

Thanks a lot for you help!

webpro:
I like how you've been digging into release-it's internals :)

My issue with this is that we would inject another life-cycle method to the core plugins just to get that script hook in-between (they're actually different concepts). This is not very pretty, and what about the next, slightly different use case... [Execution order](https://github.com/release-it/release-it/blob/master/docs/plugins/README.md#execution-order) is not always the best part of plugins here, but let's give it some more thought first. Happy to hear any ideas!

webpro:
Still, I'm a bit surprised the APK builder requires the bump to be committed. I haven't seen a situation yet where a build is not suitable from (`scripts.beforeBump` or) `scripts.beforeStage`. Do you have a repo or example you can share?

pierreroth64:
Hi @webpro, some thoughts:

- to answer your question, yes I absolutely need the bumped version to be committed to version control **before** calling the build process. Why ? because the resulting hash is included in the produced package. This is now common practice in critical domains (such as Medical, Automautive, Avionics, ...) to have the version control revision number (either a hash for git or mercurial or an id for svn) embedded in the produced package/binary/etc... (whatever the output)
- the general software build process (as described in https://en.wikipedia.org/wiki/Software_build) has a `build` step (which may have sub-steps as `compile` &`link` for C/C++ for example). This step is hidden for interpreted languages such as JS, Python, Ruby, etc... But providing such a step in `release-it` would make is available for a wider range of devs.
- even when no build step is need, one would perhaps need to create a `.tar.gz` of its documentation and may desire to include the hash in the artifact final name such as: `<my-package>-doc-45ef6d3-v1.3.0.tar.gz` and such a requirement needs something to happen between the version commit and the release push to GH.

Using `npx`the C/C++ devs (or more generally the ones that need to actually build something from sources), could run something like: `npx release-it --scripts.build='make -j4`. Providing this as a hook script would enable them to use this great tool without digging into js and writting a plugin, etc...

Working on how to reorder plugin execution seems a bit complex to me. We would then need a way to register a plugin at a given place in the plugins chain... I think that addressing this `build` step would solve a lot of use cases easily. Implementing such a step with a simple hook would avoid a lot of coupling and keep devs (that have other problems) away from the js wonderfull dev world ;)

pierreroth64:
🤔 ... hmmmm... it appears that this `"phase" or "step"` should be called `build` from what is well known by developers and not `prepareRepo` as I proposed it.

webpro:
For some context... release-it once had a `buildCommand` script hook, which is now basically the `scripts.beforeStage` hook (e.g. to bump files, update the changelog, etc. to be included in the release commit). Depending on the configuration of things like the `.gitignore` file, `git.addUntrackedFiles`, `github.assets` (and e.g. `.npmignore`, `package.json#files`, `npm.publishPath`) one can choose what build artefacts should be included/excluded precisely. I've decided to name the hooks after the moment in the process, as basically the whole purpose of this tool is to "build" and "release" things, the various hooks are meant to provide flexibility of when exactly things should happen (and `release` consequently is the "main" hook). 

For the record, also in interpreted languages, source code often code needs to be compiled/transpiled.

Generating a filename based on the next `version` from a script hook is supported in release-it (e.g. `scripts.afterBump: "tar -czvf ${name}-${version}.tar.gz"`). The `${latestVersion}` and `${latestTag}` variables are also available (yet not `${hash}`).

In summary, whether it should be possible to build something is not the point, it's more about when exactly this needs to be done in this case. I like release-it to be a "universal" release tool, like you also envision.

What we have here, is a particular situation in which that commit hash is required **and** build artefacts _based on this hash_ should be included with the GitHub release assets (maybe surprisingly, but haven't had/seen the need for this before).

I have thought before that the VCS part should probably be separated more from the other (core) plugins (while still supporting to use a different VCS plugin for e.g. Mercurial), and now you have a practical need for this. My intention now is to split them up a bit and take it from there. This does involve a bit more work than this PR, but we will get there! Again, ideas welcome. Thanks a lot for sharing your feedback!

webpro:
@pierreroth64 I've pushed a different solution, which I hope to be pretty flexible. The idea is to add a new "hooks" feature, next to the existing "scripts" hooks (and with idea to eventually phase out "scripts" and align both the script hooks and plugin life cycle methods).

I think the new test in https://github.com/release-it/release-it/commit/d557cf79a7f5e6480e62e1ea36c7bf7d4fd7c07e hopefully shows the idea best: one can now add `before/after:namespace:hook` to the `hooks` config (much like the `scripts` config). So in your use case you could hook in at `hooks["after:git:release"]` or `hooks["before:github:release"]`.

What do you think? Could you please try `npm install release-it@next` and see if this works for you?

pierreroth64:
Wow nice work @webpro ! Will try that asap. Two remaining questions:

1. Are the "new" hooks available from command line? For example: `npx release-it@next prepatch --hooks.after.git.release="echo AFTER RELEASE"` 
2. While steps are running they gather information which can be later used by the following steps. Is there a ${hash} var that the hook can use ? For example: `npx release-it@next prepatch --hooks.after.git.release="my-build-script.sh --git-hash=${hash}"`

webpro:
1. Yes, but use colons like `--hooks.after:git:release="echo after release"`
2. Not yet, but after our discussion I think it's a good idea to add it. Will be there soon (or feel free to open a PR). What is [documented here](https://github.com/release-it/release-it#scripts) is available.

Btw, remember to either escape `--hook="\${name}"` variables from the command line or use single quotes (`--hook='${name}'`).

pierreroth64:
Ok tested it and seems to work fine with the use case: `./node_modules/.bin/release-it prepatch --hooks.after:git:release="npm run build"` where `npm run build` calls a `./build.sh` script containing:

```bash
#!/usr/bin/env bash

echo 'Building software...'
exit 0
```
A comment:
- when the hook script exits with a non-zero code, `release-it` stops (which is fine) and displays the error and all logs to the console (_'Building software...'_)
- when the hook script exits with a zero-code, `release-it` continues (which is fine) **but** does not display all logs to the console (_'Building software...'_ is missing) which is a strange behaviour. The developer will probably want the build script to be verbose to be able to check settings, cli args, etc...

webpro:
You can add `--verbose` for this. This will also make release-it itself verbose, though (but please file a different issue if needed).

pierreroth64:
Thanks a lot for this feature, will use it *now* with `npx release-it@next` for my JS *and* non-JS repos 🎉

PR #532, opened 2019-05-29
Add proxy support for GitLab plugin

BinaryProvider:
@webpro, here's a new (hopefully clean) PR for the GitLab proxy support. Sorry for the inconvenience. :)

### DIFF
--- a/lib/plugin/gitlab/GitLab.js
+++ b/lib/plugin/gitlab/GitLab.js
@@ -2,6 +2,7 @@ const fs = require('fs');
 const path = require('path');
 const got = require('got');
 const _ = require('lodash');
+const globalTunnel = require('global-tunnel-ng');
 const globby = require('globby');
 const FormData = require('form-data');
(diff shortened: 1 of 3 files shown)


webpro:
Thanks @BinaryProvider. If I understand correctly, [global-tunnel-ng](https://github.com/np-maintain/global-tunnel) adds a global proxy, not just for GitLab. Also, it says "does not work for Node 11" (while release-it itself currently aims to support v8 and up).

Since release-it uses [got](https://github.com/sindresorhus/got), I think we should look into https://github.com/sindresorhus/got#proxies and find a way to configure that.

Feel free to adjust this one, or open a new PR with a solution that can be configured and is effective for release-it as a whole (e.g. including the GitHub plugin). This would be a breaking change (it would remove the `proxy` option from `github`), but that's OK.

Alternatively, you could extend (or fork) the GitLab plugin, and add the proxy in there (to not affect other folks using release-it).

PR #562, opened 2019-10-07
Catch block on templates to assist in debugging

paullryan:
I spent way too long tracking down an error in a template I had written pre-v10 that broke with an error message that didn't point to my bad template in my config.


webpro:
Sorry to hear this, @paullryan! Going to release this fix now. Hopefully, release-it will save you time eventually :)

paullryan:
@webpro no worries, I just figured since it hit it I was unlikely the not the only one. Also release-it is my go to release management tool so 👍

PR #563, opened 2019-10-08
Update README.md

ericcarboni:
Fix hyperlink in README


webpro:
Nice catch @ericcarboni! Thanks

PR #583, opened 2019-12-31
Add auto-changelog template file to be used directly from the repository

ruzickap:
Storing the auto-changelog template as a separate file may help other to use it easily form this repository:

```
auto-changelog --stdout --commit-limit false --unreleased --template https://raw.githubusercontent.com/release-it/release-it/master/docs/recipes/changelog-compact-template.hbs
```

This is handy because some people do not want to maintain it's own auto-changelog template and are happy with the default one.


webpro:
Great idea, thanks @ruzickap!

PR #587, opened 2020-01-12
Move both auto-changelog recipes to the `/conf` directory

ruzickap:
Move both `changelog-compact-template.hbs` and `keepachangelog.hbs`
auto-changelog recipes to the `/conf` directory, because they are more
like config files of auto-changelog than part of the documentation.
Change the documentation file related to auto-changelog to show examples
using recipes from this repository (`--template https://...`).


webpro:
Thanks @ruzickap! I really appreciate this. One minor thing: could you please remove the `-template` suffix from the filename(s), as I think "hbs" already implicates they're templates files, right?

webpro:
Thanks again!

PR #599, opened 2020-02-13
Fix links to documentation

walokra:
Also fixed typo of GitHub when documentation is about GitLab.


webpro:
Thanks, @walokra! Much appreciated.

Issue #602, opened 2020-02-24
Verify as collaborator with GitLab on init doesn't work for inherited members of a project


walokra:
Seems that the changes introduced in [Authenticate and verify as collaborator with GitLab on init](https://github.com/release-it/release-it/commit/923a56cb55625874f854555022038776e023f0fa ) doesn't return collaborators correctly if project members are defined in group level and are inherited members to the project.

The check is made to project members (GET /projects/:id/members) endpoint which doesn't return all members. The endpoint should be "GET /projects/:id/members/all" which returns all members of a project including inherited members. Or use "GET /projects/:id/members/all/:user_id" directly.

After we updated from 12.4.3 to >=12.5.0 the release-it gives the "is not a collaborator for" error when we run it on GitLab CI pipeline. Our users are defined as group owners and inherited to the project.

webpro:
Thanks, @walokra and @ benediktberger!

Could you please try `release-it@12.6.2-next.0` and verify it works for inherited members as well?

benediktberger:
Ah sorry, didn‘t see this issue although I browsed the issue overview 🤦‍♂

webpro:
Actually I was really happy to see a PR for this, thanks for that, Benedikt. But while I was looking at the options I found an alternative which should be faster. Does it work for you now when you `npm install release-it@next`?

PR #603, opened 2020-02-24
feat(gitlab): query inherited members from projekt

benediktberger:
Hi,

on our GitLab instance we are managing our users on group level. I noticed that using your query only direct members of the project are returned which results in an error for my current user when releasing a project.
Luckily the GitLab API doc says that we can query inherited members as well, see https://docs.gitlab.com/ee/api/members.html#list-all-members-of-a-group-or-project-including-inherited-members.

I don't see any disadvantages querying the inherited members as well, whats your opinion on this?

Best regards,

Benedikt


webpro:
Thanks @benediktberger! Now I know about `members/all/[id]` I prefer that one, please see #602.

PR #609, opened 2020-02-27
Don't include repo on push if branch has upstream

### LINKED ISSUE #608: Can't release from branch that is tracking non-origin remote
I have a `master` branch configured like so:
```
[branch "master"]
	remote = upstream
	merge = refs/heads/master
	rebase = true
```
When I try to release from that branch, it fails with:
```
ERROR fatal: You are pushing to remote 'origin', which is not the upstream of
your current branch 'master', without telling me what to push
to update which remote branch.
```
So the code is explicitly trying to push ([see here](https://github.com/release-it/release-it/blob/644050a76a51a27e67871becdad09b792ba21c2d/lib/plugin/git/Git.js#L176)) to `origin` remote which is not the remote the branch is tracking. I know I can configure remote with `git.pushRepo` but doing that would be wrong as that would affect every developer on the project and there is no guarantee they have set up their git repo the same way.

Since `release-it` already (by default) enforces that branch is tracking upstream branch, I don't see why it's explicitly passing upstream in this case. I think it would be enough to just invoke `git push` alone.

rchl:
When the branch is tracking upstream branch there should
be no need to explicitly state the origin name. Especially as
that might not be the one that matches tracking information.

Resolves #608

### DIFF
--- a/lib/plugin/git/Git.js
+++ b/lib/plugin/git/Git.js
@@ -170,8 +170,13 @@ class Git extends GitBase {
     } else if (!(await this.hasUpstreamBranch())) {
       // Start tracking upstream branch (`pushRepo` is a name if set)
       upstream = `--set-upstream ${pushRepo || upstream} ${await this.getBranchName()}`;
-    } else if (pushRepo && !invalidPushRepoRe.test(pushRepo)) {
-      upstream = pushRepo;
+    } else {
+      if (!pushRepo || pushRepo === 'origin') {
+        // Push to tracking upstream branch
+        upstream = '';
+      } else if (!invalidPushRepoRe.test(pushRepo)) {
+        upstream = pushRepo;
+      }
     }
     return this.exec(`git push ${args || ''} ${upstream}`);
   }
--- a/test/git.js
+++ b/test/git.js
@@ -165,7 +165,22 @@ test.serial('should push to origin', async t => {
   const gitClient = factory(Git);
   const spy = sinon.spy(gitClient.shell, 'exec');
   await gitClient.push();
-  t.is(spy.lastCall.args[0], 'git push  origin');
+  t.is(spy.lastCall.args[0], 'git push  ');
+  const actual = sh.exec('git ls-tree -r HEAD --name-only', { cwd: bare });
+  t.is(actual.trim(), 'file');
+  spy.restore();
+});
+
+test.serial('should push to tracked upstream branch', async t => {
+  const bare = mkTmpDir();
+  sh.exec(`git init --bare ${bare}`);
+  sh.exec(`git clone ${bare} .`);
+  sh.exec(`git remote rename origin upstream`);
+  gitAdd('line', 'file', 'Add file');
+  const gitClient = factory(Git);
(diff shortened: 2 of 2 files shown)


rchl:
I can have a look at failing tests and add new one if you think it makes sense.

webpro:
Thanks @rchl! Looks good at first glance, but please also check and fix the tests. Would be much appreciated! And might even surface a bug.

rchl:
After looking more into it, I feel like there is no perfect way of implementing that **without** breaking backwards compatibility.

I think ideal solution would be to:
 - set default value of `git.pushRepo` to an empty string
and then:
 - if `git.pushRepo` is set by the user, use that
 - if `git.pushRepo` has default value:
   a) if branch is tracking upstream - push to that
   b) if branch is not tracking upstream - push to `origin`

With current logic where `git.pushRepo` is set to `origin` by default, it's impossible to know if that is just the default value or user really wants to force the `origin` repo`.

rchl:
Before I go further with it, I need to solve a mystery.
One of the tests is failing for me without me even doing any changes:
```
git › should push to remote name (not "origin")

  /private/var/folders/d8/c7ngbjld0px_b42v6xy_d4s00000gn/T/release-it-5596D7fjiaEgt5o0/lib/shell.js:49

  Rejected promise returned by test. Reason:

  Error {
    message: `fatal: You are pushing to remote 'upstream', which is not the upstream of␊
    your current branch 'master', without telling me what to push␊
    to update which remote branch.␊
    `,
  }
```
The error makes sense but I don't get why it's not triggering on Jenkins. Does Jenkins has some different global git settings that make this not trigger?
I've tried locally with both git `2.25.0` and `2.21.1 (Apple Git-122.3)` and both trigger error while on Jenkins `2.24.0` is used and it doesn't trigger this error...

rchl:
Fixed tests and updated logic.

This change could be considered breaking since there is one case where behavior will differ.
When:
 - a branch is tracking upstream
 - `git.pushRepo` is at default `origin` value

Previously that would trigger `git push origin` thus always push to origin even if tracking information were pointing to another remote.
With those changes, in the same situation, `git push` will be triggered and will push to upstream branch.

I think current behavior makes more sense but there is a very tiny chance someone might be affected by this change.

webpro:
Yeah, I've had similar challenges with `npm.tag: latest` and such (when setting a default value in release-it, unable to see if the user explicitly has set this, or not at all).

What might be interesting here is that I'm pretty far with work on [v13](https://github.com/release-it/release-it/commits/v13) (mainly about updating dependencies and dropping support for Node v8), so perhaps it makes sense to merge your ideal solution (I agree!) to this branch.

We could still merge your current work into v12, but I prefer not to introduce this potentially breaking change (basically the same reasoning I bumped to https://github.com/release-it/release-it/blob/master/CHANGELOG.md#v12).

Would you be OK with you implementing your ideal solution, merge into v13, and I'll release v13 happen soon?

rchl:
Sure :)

rchl:
I'm fine with skipping this in v12.

PR #614, opened 2020-03-04
Try to use branch's upstream remote when releasing

rchl:
Will try to get current branch upstream remote before falling back to origin.

### DIFF
--- a/lib/plugin/GitBase.js
+++ b/lib/plugin/GitBase.js
@@ -50,11 +50,20 @@ class GitBase extends Plugin {
     return !_.includes(remoteUrlOrName, '/');
   }
 
-  getRemoteUrl() {
-    const remoteNameOrUrl = this.options.pushRepo || 'origin';
+  async getRemoteUrl() {
+    const remoteNameOrUrl = this.options.pushRepo || (await this.getBranchUpstreamRemote()) || 'origin';
     return this.isRemoteName(remoteNameOrUrl)
       ? this.exec(`git config --get remote.${remoteNameOrUrl}.url`, { options }).catch(() => null)
-      : Promise.resolve(remoteNameOrUrl);
+      : remoteNameOrUrl;
+  }
+
+  async getBranchUpstreamRemote() {
+    let remote = null;
+    const branchName = await this.exec('git rev-parse --abbrev-ref HEAD', { options });
+    if (branchName) {
+      remote = await this.exec(`git config --get branch.${branchName}.remote`, { options }).catch(() => null);
+    }
+    return remote;
   }
 
   fetch() {
--- a/lib/plugin/GitBase.js
+++ b/lib/plugin/GitBase.js
@@ -51,19 +51,23 @@ class GitBase extends Plugin {
   }
 
   async getRemoteUrl() {
-    const remoteNameOrUrl = this.options.pushRepo || (await this.getBranchUpstreamRemote()) || 'origin';
+    const remoteNameOrUrl = this.options.pushRepo || (await this.getRemote()) || 'origin';
     return this.isRemoteName(remoteNameOrUrl)
       ? this.exec(`git config --get remote.${remoteNameOrUrl}.url`, { options }).catch(() => null)
       : remoteNameOrUrl;
   }
 
-  async getBranchUpstreamRemote() {
-    let remote = null;
-    const branchName = await this.exec('git rev-parse --abbrev-ref HEAD', { options });
-    if (branchName) {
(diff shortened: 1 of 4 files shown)


rchl:
The tests are passing on Travis but locally I get failure in `github: should release to enterprise host`. It seems to be due to my local branch tracking `rchl` remote instead of `origin`. So I will have to look into that to make things more robust in testing and not rely on repo remote.

lib/plugin/GitBase.js
```diff
@@ -50,11 +50,20 @@ class GitBase extends Plugin {
...
+    const remoteNameOrUrl = this.options.pushRepo || (await this.getBranchUpstreamRemote()) || 'origin';
     return this.isRemoteName(remoteNameOrUrl)
       ? this.exec(`git config --get remote.${remoteNameOrUrl}.url`, { options }).catch(() => null)
-      : Promise.resolve(remoteNameOrUrl);
+      : remoteNameOrUrl;
+  }
+
+  async getBranchUpstreamRemote() {
+    let remote = null;
+    const branchName = await this.exec('git rev-parse --abbrev-ref HEAD', { options });
```
webpro: Please reuse `this.getBranchName`

lib/plugin/GitBase.js
```diff
@@ -50,11 +50,20 @@ class GitBase extends Plugin {
...
+    if (branchName) {
+      remote = await this.exec(`git config --get branch.${branchName}.remote`, { options }).catch(() => null);
```
webpro: This also looks like a candidate command for reuse.

lib/plugin/GitBase.js
rchl: I would need to move `getBranchName` to this class, I believe, since GitBase can be extended by other classes than `Git` and then `getBranchName` wouldn't be available.

rchl: BTW. It's kinda unclear what should be in Git and what in GitBase. I would guess that GitBase should have all pure-git stuff that is not related to release-it functionality itself but then Git has methods like `this.hasUpstreamBranch()` and `this.tagExists()` that, according to that understanding, would fit better in GitBase.

rchl: Done some changes. Let me know if that's what you had in mind. You can be picky about naming and all as I was contemplating those a bit.

rchl:
Will add test(s) when I have more time.

lib/plugin/GitBase.js
webpro: Thanks, this is what I had in mind.

Regarding the classes. Everything is now a plugin, so all release-it does is executing methods on those (i.e. there is no "release-it functionality"). GitBase is the base class for the Git plugin, but also the base for the GitRelease class, which in turn is the base class for both the GitLab and Github plugins. I just wanted to put common functionality in a base class for reuse, I guess.

rchl: What I had in mind by saying "release-it functionality" is for example: `hasUpstreamBranch` or `tagExists` being pure-git functionality, versus `isRequiredBranch` being related to the package's functionality. I would think that former two could be in GitBase, if following that logic.

webpro: Right, I see. One could argue that e.g. `isRequiredBranch` and `hasUpstreamBranch` is specific Git-plugin functionality, as the other plugins don't need it. E.g. when doing a GitHub release, this is not used/required.

Btw, I now see that `tagExists` is not even used at all, so I can remove it :)

webpro:
Thanks @rchl! I merged your PR, but I refactored one test into a more concise one, I hope you'll agree with this. I just released [v13.0.0-beta.3](https://github.com/release-it/release-it/releases/tag/13.0.0-beta.3), including both your PRs, thanks again!

rchl:
Sounds good.
I was also thinking that it should be mentioned in breaking changes that releasing to github/gitlab will use branch's upstream by default instead of `pushRepo || 'origin'`

webpro:
Isn't it more of a fix than a breaking change?

> Changed default value of `git.pushRepo` from `"origin"` to `""`. Changes behavior when branch has tracking information and `git.pushRepo` is set to default value. Previously would push to `origin`, now will push to tracked branch.

Thinking about this now: it's definitely changed behavior, but in what situation does it **break** a certain configuration/setup? It was never _intentional_ when it pushed to `origin` (like your own use case), this was a bug.

rchl:
Similar to my previous change, this will change behavior when branch is tracking upstream that is named something else than "origin". So it would break when someone relied on it always pushing to origin, even when branch was tracking something else. But yeah, I'm not sure if that would be considered a bug or a feature. ;)

PR #621, opened 2020-03-17
Support GitHub Actions

### LINKED ISSUE #618: GitHub Actions - Could not authenticate with GitHub using environment variable "GITHUB_TOKEN"
I'm getting error `Could not authenticate with GitHub using environment variable "GITHUB_TOKEN"` when running `release-it` on GitHub Actions.

The GitHub API returns: `Resource not accessible by integration`, because the GitHub integrations tries to fetch the user related to the token, in GitHub actions, the token provided is not associated to any users, so there's a few options in order to support GH Actions with release-it:
- allow to skip this validation
- find other ways to validate that the token is valid
- ??

Action being run is
```
      - name: Checkout
        uses: actions/checkout@v2
        with:
          fetch-depth: 1
      - name: New version detected, release package
        run: |
          yarn release
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

In the mean time I'll need to use a "personal token" but I'd prefer to use the GITHUB_TOKEN provided by GH actions which would make the releases published by `github-actions` instead of `wfortin`

Expected:
image

But right now with personal token, it gives this:
image

vinsidious:
## Overview
This PR adds support for running `release-it` on GitHub Actions and fixes #618.

## Background
When running on GitHub Actions, GitHub injects the `GITHUB_TOKEN` environment variable automatically. This token differs from typical GitHub user tokens in the following ways:
* It has a [predefined list of permissions](https://help.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token#permissions-for-the-github_token)
* It is unique to the job being run and expires after 60 minutes
* It is not tied to a user

Because of this, we need to skip the user authentication and collaborator checks when we're running on GitHub Actions.

#### Example of this running successfully
Screen Shot 2020-03-17 at 1 12 49 PM

### DIFF
--- a/lib/plugin/github/GitHub.js
+++ b/lib/plugin/github/GitHub.js
@@ -32,6 +32,14 @@ class GitHub extends Release {
 
   async init() {
     await super.init();
+
+    // If we're running on GitHub Actions, we can skip the authentication and
+    // collaborator checks.
+    if (this.global.isGitHubAction) {
+      this.setContext({ github: { username: process.env.GITHUB_ACTOR } });
+      return;
+    }
+
     if (!(await this.isAuthenticated())) {
       throw new GitHubClientError(
         `Could not authenticate with GitHub using environment variable "${this.options.tokenRef}".`
--- a/lib/plugin/github/GitHub.js
+++ b/lib/plugin/github/GitHub.js
@@ -35,7 +35,7 @@ class GitHub extends Release {
 
     // If we're running on GitHub Actions, we can skip the authentication and
     // collaborator checks.
-    if (this.global.isGitHubAction) {
+    if (this.isGitHubAction) {
       this.setContext({ github: { username: process.env.GITHUB_ACTOR } });
       return;
     }
--- a/lib/plugin/github/GitHub.js
+++ b/lib/plugin/github/GitHub.js
@@ -34,8 +34,8 @@ class GitHub extends Release {
     await super.init();
 
     // If we're running on GitHub Actions, we can skip the authentication and
-    // collaborator checks.
-    if (this.isGitHubAction) {
+    // collaborator checks. Ref: https://bit.ly/2vsyRzu
+    if (process.env.GITHUB_ACTION) {
       this.setContext({ github: { username: process.env.GITHUB_ACTOR } });
       return;
     }
--- a/lib/config.js
+++ b/lib/config.js
@@ -106,11 +106,6 @@ class Config {
     return Boolean(this.options.ci);
   }
 
-  get isGitHubAction() {
-    // Ref: https://bit.ly/2vsyRzu
-    return Boolean(process.env.GITHUB_ACTION);
-  }
-
   get isCollectMetrics() {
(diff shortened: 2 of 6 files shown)


webpro:
Thanks, @vinsidious! What's the reason you made the variable part of the global config? I think we can/should keep its scope entirely within the GitHub plugin.

vinsidious:
@webpro I actually went back and forth with that decision. I think my rationale was: if the execution environment was within a GitHub Action, that's on the global level and sort of pervades every other action/plugin. 

That being said, you obviously have way more context here and I also agree with you. I'll make the change 👍

webpro:
Makes total sense. But I think e.g. the npm plugin is not interested in this. If this ever changes we can of course move it up again.

vinsidious:
@webpro Hopefully this is what you had in mind? Please let me know if this isn't the right approach

lib/plugin/github/GitHub.js
```diff
@@ -32,6 +32,14 @@ class GitHub extends Release {
 
   async init() {
     await super.init();
+
+    // If we're running on GitHub Actions, we can skip the authentication and
+    // collaborator checks.
+    if (this.isGitHubAction) {
```
webpro: If we basically put `process.env.GITHUB_ACTION` right here, we don't need to add anything to the `GitBase` class (which is also the base class for the Git and GitLab plugins), while keeping our tests more clean/real. In the test for the Github Action you can actually set this environment variable to verify it works properly based on the actual "trigger" (the env var, not the option), keeping the others tests unaffected.

vinsidious: Good point! I will have to manually set that env var to be undefined for all the other tests, though since they will in fact be running on GitHub Actions 😂

webpro: Ah LOL, indeed

webpro: Forgot to mention, here is where other env vars are (un)set: https://github.com/release-it/release-it/blob/master/test/util/setup.js

vinsidious:
@webpro Looks like the tests are passing but I do want to call out that I've noticed them being a _little_ iffy. I imagine it has something to do with ava's concurrency but appending `.serial` to the suspect test cases didn't seem to help very much.

webpro:
@vinsidious Do you recall what kind of errors you were seeing? I'm on macOS, so if you're using something different this might be useful info.

vinsidious:
I'm on macOS as well, but the [logs from the failed tests](https://github.com/release-it/release-it/pull/621/checks?sha=848307bed13ba4c347c00fbfe82b0fb88e678f45) on GitHub are the same as the errors I was getting locally. I was able to consistently repro by commenting/uncommenting the line that sets `process.env.GITHUB_ACTION = 'foo-123'`

webpro:
Thanks a lot, @vinsidious! Merged and released as [v13.2.0-beta.0](https://github.com/release-it/release-it/releases/tag/13.2.0-beta.0).

PR #634, opened 2020-03-30
[bug fix] tagName being ignored

zshannon:
Hi folks, here's a bug fix for `tagName` config which is being ignored in `v13.5.0` :)


webpro:
That's a subtle bug. Thanks, @zshannon!

webpro:
In [v13.5.1](https://github.com/release-it/release-it/releases/tag/13.5.1)

PR #639, opened 2020-04-06
fix gitlab "is-collaborator" check (adjust API call)

driiftkiing:
I have found a bug, when trying to check if the user is a collaborator.

\
Please feel free to tell me, if I have to change the description. 😄

### DIFF
--- a/lib/plugin/gitlab/GitLab.js
+++ b/lib/plugin/gitlab/GitLab.js
@@ -65,13 +65,21 @@ class GitLab extends Release {
   async isCollaborator() {
     if (this.global.isDryRun) return true;
     const { id } = this.getContext('gitlab');
-    const endpoint = `projects/${this.id}/members/${id}`;
+    const endpoint = `projects/${this.id}/members/all/${id}`;
     try {
       const user = await this.request(endpoint, { method: 'GET' });
       return user && user.access_level >= 30;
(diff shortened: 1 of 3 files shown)


driiftkiing:
its seems like, the api `projects/:id/members/all/:id` was introduced in [v12.4.0](https://docs.gitlab.com/ee/api/members.html#get-a-member-of-a-group-or-project-including-inherited-members)

is it possible to add a version check? its not always possible to having the gitlab running at the latest version

webpro:
Another idea is to retry in the `catch` block of the current `isCollaborator` method. Then we don't need a version check, we just try again if the `/members/all/${id}` request fails with a specific error code/response, and then use the `/members/${id}` path instead. What do you think?

driiftkiing:
Yeah, was also thinking about that, but had the idea with the version, because then it's clear why there is choosen another API to get the result.

But we can go also with the `try/catch` block, if you want that.

webpro:
Please also check the type of error. I hope there is a difference between e.g. "endpoint not found" and any other error, in which case we don't need to make an extra request.

webpro:
Thanks @driiftkiing! It's in [v13.5.2](https://github.com/release-it/release-it/releases/tag/13.5.2).

PR #646, opened 2020-04-20
docs(ci): added GitLab CI configuration

lucbas:
Got it working with the described config. If helpful, can be added to the repo. We could probably change the order of the new section or rename it.


webpro:
Thanks @lucbas! I've merged and made minor edits to it. Much appreciated!

PR #659, opened 2020-05-01
Add support for reusing and editing an existing release

Page-:
This is useful if you want to draft a release and upload multiple assets to it before finally publishing it, eg because the assets are generated on different OSes


webpro:
Thanks, interesting idea. How is this feature supposed to be used? Add `--github.reuseRelease`, and hardcode the version (`increment`)? Also `--no-git` I guess?

I'm asking because it's a limitation in release-it now, but with `--draft` this can be done only once, right? Because only `publishRelease` actually updates the release, but always with `draft: false`.

Page-:
The way I've been using it so far is as
```
# Upload asset 1
release-it \
	--no-npm \
	--no-increment \
	--no-git \
	--github.draft=true \
	--github.release=true \
	--github.tagName="$tagName}" \
	--github.releaseName="${releaseName}" \
	--github.assets=$ASSET1 \
	--github.reuseRelease=true
# Upload asset 2
release-it \
	--no-npm \
	--no-increment \
	--no-git \
	--github.draft=true \
	--github.release=true \
	--github.tagName="$tagName}" \
	--github.releaseName="${releaseName}" \
	--github.assets=$ASSET2 \
	--github.reuseRelease=true
# Publish as non-draft
release-it \
	--no-npm \
	--no-increment \
	--no-git \
	--github.release=true \
	--github.tagName="$tagName}" \
	--github.releaseName="${releaseName}" \
	--github.reuseRelease=true
```
and it seems to work nicely with a draft release that accumulates the assets before the final publish.

One issue I hit is that it'll fail if you try to upload a later asset with the same name but the only use case I can think of for that is to replace a broken asset which would already require manual intervention so I've left it for now - I might look at it if I find a need as it should just be a case of deleting matching assets before uploading

webpro:
Thanks @Page-, you inspired me to make some structural changes and start supporting updating existing releases. Also the github plugin is refactored quite a bit. I've just released [v13.6.0-beta.0](https://github.com/release-it/release-it/releases/tag/13.6.0-beta.0) with this.

Would you mind giving it a go? You can install using e.g. `npm install release-it@beta`.

You can use it as in your own examples, except that you don't need the `--github.reuseRelease` flag, this is inferred from the `--no-increment` argument.

One more note, you can use e.g. `--github.draft` for flags (instead of `--github.draft=true`).

Happy to hear your thoughts and findings!

Page-:
I tried this with `npx release-it@beta` @webpro and it didn't seem to work, it just created two releases. I also checked the version was correct with
```
$ npx release-it@beta --version
v13.6.0-beta.0
```

I also found that using  `--github.draft=true` failed with
```
ERROR 422 Unprocessable Entity (Invalid request.

For 'properties/draft', "true" is not a boolean.)
````
which isn't a big problem as I just switched it over to `--github.draft` but it'd be nice to improve the error message

**Comment by webpro** [comment]
@Page- thanks you for testing! I found that GitHub's API method `getReleaseByTag` does not support drafts. So things should work better now. I ran a series of commands and the `draft` status can be changed and assets can be added:

```
# start by creating a release
release-it --github.release --github.draft --no-npm.publish

# add asset
release-it --github.release --github.draft --no-npm.publish --no-increment --github.assets=file --no-git

# add another asset
release-it --github.release --github.draft --no-npm.publish --no-increment --github.assets=file2 --no-git

# remove draft status
release-it --github.release --no-npm.publish --no-increment --no-git
```

Any change you could verify? You can use `npx release-it@beta` again. Thank you in advance!

webpro:
This was just released in [v13.6.0](https://github.com/release-it/release-it/releases/tag/13.6.0), so closing this issue. Thanks again, @Page-! Interesting idea and useful for future features.

Page-:
Thanks @webpro, I've tested the latest version and it looks to be working well

PR #667, opened 2020-05-18
Only update a release with a matching tag name

Page-:
Carrying on from #659 I tested a bit further and the current situations means that running
```
release-it \
	--no-npm \
	--no-increment \
	--no-git \
	--github.draft \
	--github.release \
	--github.tagName="pr1-draft" \
	--github.releaseName="pr1 draft" \
	--github.preRelease \
	--github.assets=${ASSETS}
release-it \
	--no-npm \
	--no-increment \
	--no-git \
	--github.draft \
	--github.release \
	--github.tagName="pr2-draft" \
	--github.releaseName="pr2 draft" \
	--github.preRelease \
	--github.assets=${ASSETS}
```
would have the second release update the first despite the tag name and release name being completely different which was unexpected behavior to me, this PR restores the behavior of my original PR where the release is only updated if the names match so that it's possible to have multiple drafts at once.

For my use case it's a draft per PR so that the release artifacts of a PR can be generated/tested and then reused on a merge, which also adds a need to be able to explicitly rename the release to go from the PR naming schema to the versioned `v1.2.3` naming scheme once a PR is merged - I'll open a separate issue for that part but the use case should help clarify things

### DIFF
--- a/lib/plugin/github/GitHub.js
+++ b/lib/plugin/github/GitHub.js
@@ -150,15 +150,19 @@ class GitHub extends Release {
 
   async getLatestRelease() {
     const { owner, project: repo } = this.getContext('repo');
+    const { latestTagName } = this.getContext();
     try {
       const options = {
         owner,
         repo
       };
       this.debug(options);
-      const response = await this.client.repos.listReleases({ owner, repo, per_page: 1, page: 1 });
-      this.debug(response.data[0]);
-      return response.data[0];
+      const response = await this.client.repos.listReleases(options);
+      const release = response.data.find(release => {
+        return release.tag_name === latestTagName;
+      });
+      this.debug(release);
+      return release;
     } catch (err) {
       return this.handleError(err, () => {});
     }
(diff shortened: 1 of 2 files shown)


webpro:
> would have the second release update the first despite the tag name and release name being completely different which was unexpected behavior to me, this PR restores the behavior of my original PR where the release is only updated if the names match so that it's possible to have multiple drafts at once.

The idea with `--no-increment` is to update the *latest* release (hence `getLatestRelease`). My idea was to have a relatively quick and easy way (both from a user and code complexity perspective) to fix the current release that's being worked on.

Your use case is different, you want to update *any* existing release. My apologies I haven't resolved this in my other PR. I will take some time soon to see how basically `--no-increment` and `--github.update` can work together, as I don't want to rush this and end up with a weird API.

webpro:
Sorry, I'm going to close this PR.

The reason is that the spirit of release-it is to do new releases, with usually multiple plugins (e.g. Git, GitHub) being active. What we want here is basically a GitHub API to modify an existing release, but in release-it, this also means "disabling all other plugins". The `--no-increment` (to update the latest) is basically as far as we can go here.

You could make a plugin for this, as all the features are available: disable other plugins, extend from GitHub class to add or modify methods, etc.

I'm happy to discuss further if you have ideas for this, because obviously it would be a great feature to have.

Issue #616, opened 2020-03-11
Error when generating release with changelog


leandromoreirati:
Hello!
I am running the release integrated with auto-changelog and it is returning the following error:

`$ npx auto-changelog --stdout --commit-limit false -u --template https://raw.githubusercontent.com/release-it/release-it/master/conf/changelog-compact.hbs
404: Not Found`

I went to check the address of the template and it does not exist, did the address change, or is this integration being discontinued?

webpro:
This file has been moved from the `conf` directory to `templates`. I tried to made a symlink to keep it backwards compatible, but obviously this didn't work. Apologies for the inconvenience!

webpro:
I did just put some proper symlinks in place, though.

PR #672, opened 2020-05-31
docs: fix changelog template URL

favoyang:
Refs https://github.com/release-it/release-it/issues/616


webpro:
Thanks, @favoyang!

Issue #707, opened 2020-10-02
Increase existing prerelease version by patch, minor, major


pstrh:
It would be nice if release-it would behave like semver if the current version is a prerelease version and the version should now be incremented by patch, minor, major.

Semver allows to increment the version from 1.2.3-next.0 to 1.3.0-next.0 directly by calling:

    semver --increment preminor --preid next 1.2.3-next.0
    // result: 1.3.0-next.0

Instead release-it will increment from 1.2.3-next.0 to 1.2.3-next.1.

    release-it preminor --preReleaseId=next
    // result: 1.2.3-next.1

    release-it minor --preRelease=next
    // result: 1.2.3-next.1

The change would be helpful if you already released a prerelease version, but then later decide to increase the version (e.g. from minor to major).

Thank you for this great project 😃👍

webpro:
Yes I agree @pstrh, never needed this myself, but can look into this soon. Or feel free to take a stab at it and open a PR :)

pstrh:
@webpro Great! I would try to help with the issue and will create a PR within the next days.

webpro:
Nice! Let me know if/when you need any help, happy to assist.

pstrh:
Released with v14.2.0

PR #709, opened 2020-10-12
Added support to increase existing prerelease version by patch, minor…

pstrh:
PR for issue #707 

I've adapted the handling of prerelease versions and added two additional tests.

Futhermore I had to change an existing test. Previously if release-it was executed with prepatch on a prerelease version only the prerelease number was increased. Now the patch version and the prerelease suffix is adapted. I think this is in line with semver but I am not sure if it will affect other use cases/users.

### DIFF
--- a/lib/plugin/version/Version.js
+++ b/lib/plugin/version/Version.js
@@ -103,12 +103,11 @@ class Version extends Plugin {
       return increment;
     }
 
-    const _isPreRelease = isPreRelease || (increment || '').startsWith('pre');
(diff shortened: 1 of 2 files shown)


webpro:
Nice fix! LGTM, I'll test a bit and merge shortly

webpro:
Thanks @pstrh! This just landed in [v14.2.0](https://github.com/release-it/release-it/releases/tag/14.2.0)

pstrh:
Cool :) Thanks @webpro

PR #710, opened 2020-10-13
added getIncrement function to Plugin.js

zoli-kasa:
I've created a getIncrement function in Plugin.js to be able to atomatically get the increment, it would be useful if it's in a changelog file for example.


webpro:
Thanks @zoli-kasa! This just landed in [v14.2.0](https://github.com/release-it/release-it/releases/tag/14.2.0)

Issue #714, opened 2020-10-26
Prerelease continuation increases patch version in CI mode


pstrh:
#707 introduced an inconsistent behaviour of interactive and CI mode in regards to prerelease continuation.

**Steps and result:**
Current version is 2.0.0-beta.1

    release-it --preRelease=rc 

results in 2.0.0-rc.0

While

    release-it --preRelease=rc --ci

results in 2.0.1-rc.0

**Expected:**
Both commands should result in 2.0.0-rc.0. With the ```--ci``` flag the patch version should not be increased.

**Cause:**
Reason for this issue is that "patch" is the default version increment in CI mode and since #707 patch/minor/major are process in conjunction with the preRelease flag.

webpro:
Closing this as the fix has been released.

PR #715, opened 2020-10-26
Fix prerelease continuation increased patch version in CI mode

pstrh:
@webpro This PR tries to solve #714 - which I've introduced via #707 

Problem was that the 'patch' default in the CI mode previously did not had an effect in a prerelease continuation scenario. With #707 the default patch increment is applied.

To solve this the prelease shortcut has to be detected first. Only if it's not the prerelease shortcut the increment can be set to 'patch' in CI mode. To be able to implement this evaluation order I added a optional isCI flag to the incrementVersion method.

I also added some additional tests to verify the fix.

### DIFF
--- a/lib/plugin/version/Version.js
+++ b/lib/plugin/version/Version.js
@@ -63,8 +63,7 @@ class Version extends Plugin {
 
   getIncrementedVersionCI(options) {
     const { isCI } = this.config;
-    options.increment = options.increment == null && isCI ? 'patch' : options.increment;
-    return this.incrementVersion(options);
+    return this.incrementVersion(options, isCI);
   }
 
   async getIncrementedVersion(options) {
@@ -93,7 +92,7 @@ class Version extends Plugin {
     return Boolean(semver.valid(version));
   }
 
-  incrementVersion({ latestVersion, increment, isPreRelease, preReleaseId }) {
+  incrementVersion({ latestVersion, increment, isPreRelease, preReleaseId }, isCI) {
     if (increment === false) return latestVersion;
 
     const latestIsPreRelease = this.isPreRelease(latestVersion);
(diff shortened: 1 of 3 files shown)


pstrh:
Seems that I've broken some tests will have a look into it tomorrow.

test/plugins.js
```diff
@@ -106,7 +106,7 @@ test.serial('should instantiate plugins and execute all release-cycle methods',
     t.is(myLocalPlugin[method].callCount, 1);
   });
 
-  const incrementBase = { latestVersion: '0.0.0', increment: 'patch', isPreRelease: false, preReleaseId: undefined };
+  const incrementBase = { latestVersion: '0.0.0', increment: undefined, isPreRelease: false, preReleaseId: undefined };
```
pstrh, line 109: @webpro Due to my change the options.increment is no longer modified via getIncrementedVersionCI. I guess that this is not a problem. But better check carefully if this change has side effects.

lib/plugin/version/Version.js
```diff
@@ -111,6 +110,10 @@ class Version extends Plugin {
       return semver.inc(latestVersion, 'prerelease', preReleaseId);
     }
 
+    if (isCI && !increment) {
```
webpro, line 113: Thanks for your PR, @pstrh! I see the issue now and like the solution. If you could please use `this.config.isCi` here directly instead of passing the argument around (and ideally also run `npm run format`) I'm happy to merge it.

(Looks like we can even remove `getIncrementedVersionCI` from this plugin then, right?)

webpro:
Thanks for the fix, @pstrh! I did some more testing and refactored a bit, this is now in [14.2.1](https://github.com/release-it/release-it/releases/tag/14.2.1).

pstrh:
@webpro Couldn't look into this until now. Thank you for finisihing and merging this PR 👍

PR #764, opened 2021-05-10
docs(ci): add alternative .npmrc command and circle ci details

lipe-dev:
This PR will, once merged, add a few extra lines of documentation. First, suggesting a new command to edit the .npmrc file for auth purposes using NPM's own config set instead of manually editing the file, and the second is a bit of clarification for Circle CI users that might find problems in the git steps of the release process.

### DIFF
--- a/docs/ci.md
+++ b/docs/ci.md
@@ -113,7 +113,7 @@ jobs:
           command: npm config set //registry.npmjs.org/:_authToken $NPM_TOKEN
 ```
 
-During the release process, your project's `package.json` will be updated to bump the version, you will need to setup
+During the release process, your project's `package.json` will be updated to bump the version. You will need to setup
 CircleCI with a non read-only SSH key pair from your Github account if you want it to be able to push that change back
 to the repo.
 
--- a/docs/ci.md
+++ b/docs/ci.md
@@ -75,8 +75,11 @@ package:
 One way to achieve this is to set the `NPM_TOKEN` in the CI environment, and from a script do:
 
 ```bash
-echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > .npmrc
+npm config set //registry.npmjs.org/:_authToken $NPM_TOKEN
 ```
+
+This will create/update the `.npmrc` file and add the token there. Ideally you should either `.gitignore` this file, 
+otherwise you will end up committing it to your repo if you are using release-it's git options.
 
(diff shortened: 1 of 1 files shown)


docs/ci.md
```diff
@@ -75,9 +75,12 @@ package:
 One way to achieve this is to set the `NPM_TOKEN` in the CI environment, and from a script do:
 
 ```bash
-echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > .npmrc
+npm config set //registry.npmjs.org/:_authToken $NPM_TOKEN
 ```
 
+This will create/update the `.npmrc` file and add the token there. Ideally you should either `.gitignore` this file, 
+otherwise you will end up committing it to your repo if you are using release-it's git options.
```
webpro, line 82: It's not entirely correct, only when using `addUntrackedFiles: true`:

```suggestion
otherwise you might end up committing it to your repo if you are using release-it's `git` options.
```

docs/ci.md
```diff
@@ -75,9 +75,12 @@ package:
 One way to achieve this is to set the `NPM_TOKEN` in the CI environment, and from a script do:
 
 ```bash
-echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > .npmrc
+npm config set //registry.npmjs.org/:_authToken $NPM_TOKEN
```
webpro, line 78: Nice! Much better.

docs/ci.md
```diff
@@ -107,9 +110,13 @@ jobs:
           at: ~/repo
       - run:
           name: Authenticate with registry
-          command: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > ~/repo/.npmrc
+          command: npm config set //registry.npmjs.org/:_authToken $NPM_TOKEN
 ```
 
+During the release process, your project's `package.json` will be updated to bump the version, you will need to setup
```
webpro, line 116: Just a tiny change to increase readability:

```suggestion
During the release process, your project's `package.json` will be updated to bump the version. You will need to setup
```

webpro (commented):
Thanks a lot, @lipe-dev! Happy to merge if you agree with my changes.

docs/ci.md
lipe-dev: True! Thanks for pointing it out.

webpro:
Thank you, @lipe-dev! Much appreciated.

PR #767, opened 2021-05-20
docs: add missing template flag to auto-changelog

v-stickykeys:
Otherwise the hook fails to update the changelog


webpro:
Thanks a lot @v-stickykeys! Appreciated.

PR #771, opened 2021-06-02
Move changelog generation to after version bump

DomiR:
Changelog providing plugins can use the incremented version for generation


rwjblue:
Generally speaking this seems like a nice change overall (to me), but I think it will break changelog generators that are looking for changes since the last tag (e.g. introspecting git history or pull requests).

webpro:
I understand the reasoning, but the conventional plugin by default disables the "version" plugin, which we need in order to prompt for the next version.

I have released v3 of the plugin, please see https://github.com/release-it/conventional-changelog/issues/17

So I'm going to close this PR, but thanks for raising the issue and taking the time to think along.

PR #772, opened 2021-06-07
feat: Adding certificateAuthorityFile option for GitLab releases

Booyaabes:
When running a private GitLab instance protected with a certificate issued by a private certificate authority, you can now specify a root CA certificate file with `gitlab.certificateAuthorityFile` option.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -4,7 +4,7 @@
     "es6": true
   },
   "parserOptions": {
-    "ecmaVersion": 8,
+    "ecmaVersion": 2018,
     "sourceType": "module"
   },
   "extends": ["eslint:recommended", "plugin:ava/recommended", "prettier"],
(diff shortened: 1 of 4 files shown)


Booyaabes:
I would like to point out that GOT does not use the certificates installed at system level (which I find quite surprising.. to say the least).
More information [here](https://github.com/sindresorhus/got#httpscertificateauthority).

webpro:
Thanks @Booyaabes! That's a nice feature. I have a question: can we also set this option once in the `got.extend` client factory (instead of with every request)? Here:

https://github.com/release-it/release-it/blob/master/lib/plugin/gitlab/GitLab.js#L21-L34

Booyaabes:
You're right, that should work. I'll give it a try.

Booyaabes:
@webpro, it works. I just add a commit.

webpro:
Thanks @Booyaabes! It's in [v14.9.0](https://github.com/release-it/release-it/releases/tag/14.9.0)

Booyaabes:
Thx to you for this great tool !

PR #774, opened 2021-06-12
Fix: example plugin reads file as string

ChocolateLoverRaj:
Before it was getting a buffer, but it was meant to get a string, so an encoding needs to be specified.


webpro:
Thanks @ChocolateLoverRaj! I thought it was the default, but it's not.

PR #785, opened 2021-06-30
feat(status): remove whitespace from front of status

DomiR:
This is an OCD issue, but `git status --short --untracked-files=no` adds a whitespace in front of every line except the first one. It would be nice to uniformly indent the changeset or like with this PR remove the front whitespace from every line of the the git status.

This PR will remove the spaces before each M:
image

### DIFF
--- a/lib/plugin/git/Git.js
+++ b/lib/plugin/git/Git.js
@@ -155,7 +155,7 @@ class Git extends GitBase {
   }
 
   status() {
-    return this.exec('git status --short --untracked-files=no', { options }).catch(() => null);
+    return this.exec('git status --short --untracked-files=no', { options }).then(status => status.replace(/^\s+/gm, '')).catch(() => null);
   }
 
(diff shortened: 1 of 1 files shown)


webpro:
Thanks for the PR, @DomiR. The issue is actually that the first space is removed, since release-it has a `trim()` for everything that comes back from `stdout`. I'm experimenting a bit with using `trimEnd()` here, and at first sight it seems to do the job properly. What do you think?

https://github.com/release-it/release-it/blob/master/lib/shell.js#L57

DomiR:
Hm weired... I thought I ran the git command by hand to verify that it misses the first space but you are right. Maybe I just console logged the output of your exec... So trimEnd should work as well and is available for node>10 👍

webpro:
I've released this in [v14.10.1](https://github.com/release-it/release-it/releases/tag/14.10.1)

PR #856, opened 2022-01-15
Fix typo.

ronyeh:


ronyeh:
Oops, I see there is another PR with the same typo fix. :-D

PR #887, opened 2022-04-06
fix: downgrade `inquirer` from 8.2.2 to 8.2.0 to avoid dropping Node 10 support

### LINKED ISSUE #886: Breaking change in 14.14.0 regarding Node support
release-it [v14.14.0](https://github.com/release-it/release-it/releases/tag/14.14.0) includes an [update](https://github.com/release-it/release-it/commit/c61c780d6627cb9706b94301d96b1527a5fed7d8) to inquirer 8.2.2 which [drops](https://github.com/SBoudrias/Inquirer.js/commit/a1da6cfc97fe29cdeb1edc8b2aafc20eeaf85dca) support below Node 12.

This prevents users from using release-it with Node 10 which is supposed to be supported:

https://github.com/release-it/release-it/blob/afd1ac88f33e5bbbfbb58da2aadee42277b068f3/package.json#L107

Example where I hit this: https://github.com/ember-cli/eslint-plugin-ember/runs/5858294599?check_suite_focus=true

> error inquirer@8.2.2: The engine "node" is incompatible with this module. Expected version ">=12.0.0". Got "10.24.1"

I would recommend reverting the inquirer update until a subsequent major release of release-it.

It also seems that the CI tests did not catch this.

bmish:
Fixes #886.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "globby": "11.0.4",
     "got": "11.8.3",
     "import-cwd": "3.0.0",
-    "inquirer": "8.2.2",
+    "inquirer": "8.2.0",
     "is-ci": "3.0.1",
     "lodash": "4.17.21",
     "mime-types": "2.1.35",


webpro:
Thanks @bmish! This was always a bit hard to test (automatically), and from your other PR I understand the tests would not even fail...

This is released in [14.14.1](https://github.com/release-it/release-it/releases/tag/14.14.1).

PR #888, opened 2022-04-06
test: fix CI test for different Node versions

bmish:
I believe CI was failing to test the different Node versions that should be supported because of this missing step.

https://github.com/actions/setup-node#usage

Related to: #886

The Node 10 test should fail for this PR until #887 is merged.


webpro:
Pretty embarrassed about this! Glad you reported this issue. Thanks for the fix!

This is part of [14.14.1](https://github.com/release-it/release-it/releases/tag/14.14.1).

PR #892, opened 2022-04-10
Strict Node.js v10 support

webpro:
Adds `engine-strict=true`, a few dependency downgrades, and a resolution (for Yarn)


bmish (approved):

bmish:
Thanks! This finally enables me to upgrade now that the Node 10 CI check passes: https://github.com/ember-cli/eslint-plugin-ember/pull/1463

Issue #863, opened 2022-01-25
github.proxy not worked?


afeiship:
## I found in `@octokit/dist-src/index.js`

```js
export class Octokit {
    constructor(options = {}) {
        // use agent, instead of `proxy`
    }
}
```
## proxy
https://github.com/release-it/release-it/blob/master/lib/plugin/github/GitHub.js#L170

```js
get client() {
    if (this._client) return this._client;
    const { proxy, timeout } = this.options;
    const host = this.options.host || this.getContext('repo.host');
    const isGitHub = host === 'github.com';
    const baseUrl = `https://${isGitHub ? 'api.github.com' : host}${isGitHub ? '' : '/api/v3'}`;
    const options = {
        baseUrl,
        auth: `token ${this.token}`,
        userAgent: `release-it/${pkg.version}`,
        log: this.config.isDebug ? console : null,
        request: {
            timeout
        }
    };

    if (proxy) {
        options.proxy = proxy;
        // so, maybe this should be worked?
        options.request.agent = new ProxyAgent(proxy);
    }

    const client = new Octokit(options);

    this._client = client;
    return client;
}
```

webpro:
Interesting. Could you please open a PR with this fix?

Maybe there has been an update in `@octokit/rest` or `@octokit/request` I've missed.

afeiship:
@webpro thx, you have fixed the bug in new version. :)
I have a template script to fix in my project,  waiting for new verison release! 

```js
// 1. Add proxy to `.release-it.json`
// 2. ./node_modules/release-it/lib/plugin/github/GitHub.js
// 3. const prompts = require('./prompts'); ---> const ProxyAgent = require('proxy-agent');
// 4. options.proxy = proxy; --->  options.request.agent = new ProxyAgent(proxy);
// 5. add proxy to package.json
```

webpro:
Sorry @afeiship! I pushed to code to test it in the GitHub Action, but forgot to release-it! It's in [v14.14.3](https://github.com/release-it/release-it/releases/tag/14.14.3)

afeiship:
@webpro thanks, I have installed it in my project, **it worked**!

PR #893, opened 2022-04-14
Fixes github.proxy not working (#863)

dgcofer:
Based on a recent change in Octokits docs[1], in order for it to use a
configured proxy you must pass another request.agent into the Octokit
constructor. This change adds the `proxy-agent` lib as suggested by
Octokits docs.

Note that with this implmentation, the `http_proxy` or `https_proxy`
environment variables alone will NOT work. Instead one must explicitly
set the value in the `github.proxy` option passed to release-it. This is
to stay inline with current behavior[2].

[1] https://github.com/octokit/octokit.js/commit/2090c85b5addf33b93dac74930312593622c08b8
[2] https://github.com/release-it/release-it/blob/14.14.2/docs/github-releases.md#proxy

### DIFF
--- a/test/github.js
+++ b/test/github.js
@@ -407,3 +407,33 @@ test('should generate GitHub web release url for enterprise host', async t => {
   );
   exec.restore();
 });
+
+test('should generate Github web release when proxy is set', async t => {
+  const options = {
+    git,
+    github: {
+      pushRepo: 'git://custom.example.org:user/repo',
+      release: true,
+      web: true,
+      host: 'custom.example.org',
+      releaseName: 'The Launch',
+      releaseNotes: 'echo It happened',
+      proxy: 'http://proxy.example.org:8181'
+    }
+  };
+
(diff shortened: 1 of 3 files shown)


test/github.js
```diff
@@ -407,3 +407,33 @@ test('should generate GitHub web release url for enterprise host', async t => {
...
+
+  await runTasks(github);
+
+  const { isReleased, releaseUrl } = github.getContext();
+  t.true(isReleased);
+  t.is(
+    releaseUrl,
+    'https://custom.example.org/user/repo/releases/new?tag=2.0.2&title=The+Launch&body=It+happened&prerelease=false'
+  );
+  t.is(github.options.proxy, 'http://proxy.example.org:8181');
```
dgcofer, line 437: Wasn't really to sure on how to test for this scenario, Ava is not my forte 😉 
I welcome any suggestions.

PR #911, opened 2022-06-16
feat(git): make commitMessage optional parameter in git commit

### LINKED ISSUE #907: Possible to tag the last commit rather than generate a new commit?
Using GitHub actions, I'd like to be able to create a `preRelease` version every merge into my project's `main` branch by tagging whatever the previous commit was with a certain version. In theory, I should be able to do something like this to achieve that : `npx release-it --preRelease --ci --git.commitMessage="" --git.commitArgs="--amend -C HEAD"`, however `release-it` currently seems to always configure a message commit even if I specify an empty message. Is it possible to be able to tag the previous commit version with `release-it` currently?

Came across release-it recently and it's been great!

blzsaa:
Due to git.commitMessage being optional t is now possible to amend previous commit instead of
creating a new one.

Example usage:
release-it minor --git.commitArgs='--amend --no-edit' --git.commitMessage=''

fix #907

### DIFF
--- a/lib/plugin/git/Git.js
+++ b/lib/plugin/git/Git.js
@@ -168,7 +168,8 @@ class Git extends GitBase {
 
   commit({ message = this.options.commitMessage, args = this.options.commitArgs } = {}) {
     const msg = format(message, this.config.getContext());
-    return this.exec(['git', 'commit', '--message', msg, ...fixArgs(args)]).then(
+    const commitMessageArgs = msg ? ['--message', msg] : [];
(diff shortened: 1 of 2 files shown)


webpro:
Thanks @blzsaa! I will merge and then minimize the test a little to only verify what we really need for this new feature.

webpro:
This is released in [v15.1.0](https://github.com/release-it/release-it/releases/tag/15.1.0)

PR #912, opened 2022-06-23
feat(cli): add new --changelog option

husseinraoouf:
Add --changelog option to cli to be able to only get the change log similar to what --release-version do but for change log


webpro:
Thanks @husseinraoouf! It's in [15.3.0](https://github.com/release-it/release-it/releases/tag/15.3.0)

Issue #920, opened 2022-07-15
Package.json exports field defines extensionless export that errors in pure ESM environments


scalvert:
The fixed export of `release/test/util` in #539 helped consumers access the test utils again (thank you!). Using an extensionless export, however, has complicated compatibility across multiple `release-it` versions. Pure ESM `release-it` plugins, when testing using `ava`, will error with extensionless sub-paths. 

https://github.com/rwjblue/release-it-lerna-changelog/runs/7347409577?check_suite_focus=true

```shell
> release-it-lerna-changelog@4.0.1 test /home/runner/work/release-it-lerna-changelog/release-it-lerna-changelog
> ava
  Uncaught exception in test.js
  Error [ERR_UNSUPPORTED_DIR_IMPORT]: Directory import '/home/runner/work/release-it-lerna-changelog/release-it-lerna-changelog/node_modules/release-it/test/util' is not supported resolving ES modules imported from /home/runner/work/release-it-lerna-changelog/release-it-lerna-changelog/test.js
  Did you mean to import release-it-lerna-changelog/release-it-lerna-changelog/node_modules/release-it/test/util/index.js?
  ✖ test.js exited with a non-zero exit code: 1
```

The Node documentation [encourages the use of extension-based exports when those exports point to a single file](https://nodejs.org/api/packages.html#subpath-exports), as is the case here. I've tested converting the package's exports to use an extension instead of a sub-path, and it ensures correct compatibility across `release-it` versions 14-latest (15.1.1). 

Change:

```json
  "exports": {
    ".": "./lib/index.js",
    "./test/util/index.js": "./test/util/index.js"
  },
```

If you'd be open to it, I'll open a PR making that change in the latest version.

webpro:
Thanks @scalvert. Agreed, just patched it in [v15.1.2](https://github.com/release-it/release-it/releases/tag/15.1.2). Since it's a test util I consider this not a breaking change.

PR #921, opened 2022-07-17
fix: Fixes exports to correctly export test utils

scalvert:
This change slightly adjusts the exports in this repo's package.json in order to expose the required test utilities to consumers.

While the alias provided in https://github.com/release-it/release-it/commit/529bc1fde988515079efdb0a8bd2e0fc7360c17e technically works for ESM consumers, it won't work for CJS consumers as that path isn't a real path on disc.

Relates to #920 and tweaks https://github.com/release-it/release-it/commit/529bc1fde988515079efdb0a8bd2e0fc7360c17e to use correct path.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
   "type": "module",
   "exports": {
     ".": "./lib/index.js",
-    "./test/util.js": "./test/util/index.js"
+    "./test/util/index.js": "./test/util/index.js"
   },
   "files": [
     "bin",


scalvert:
cc @webpro - happy to discuss!

scalvert:
On quick reflection, what I said above is not correct. This _will_ work with CJS consumers as they'll just ignore the exports field. 

I suppose the compelling reason would be to keep the paths consistent with real paths on disc, but I'm fine if you ignore this PR too!

scalvert:
To be very clear, if we use the following path:

```js
import { factory, runTasks } from 'release-it/test/util.js';
```

This will work fine for ESM consumers since that path is aliased via `exports`, which they will observe. It won't work for CJS consumers since that's not a real path on disc, and CJS doesn't pay attention to the `exports` field.

Sorry if I'm restating the obvious :)

scalvert:
@webpro just pinging on this again. Let me know if there's anything I can do to help!

webpro:
Alright, let's play it safe :) Thanks @scalvert!

PR #922, opened 2022-07-20
Updates README with new plugin package names

scalvert:
We've moved two plugin packages under a new org, `@release-it-plugins`, and renamed them. This PR updates the README to reflect those renames.


webpro:
Nice, thanks

PR #928, opened 2022-08-07
Security issue fix: GHSA-vp56-6g26-6827

nbouvrette:
Fixes https://github.com/advisories/GHSA-vp56-6g26-6827

Still need to be compiled/tested and a new version released.

The node-fetch change is minor and should not have any impact. See: https://github.com/node-fetch/node-fetch/pull/1611/files


nbouvrette:
@webpro just wondering if you had a chance to look at this PR?

PR #958, opened 2022-12-06
fix: use spec'ed formdata

jimmywarting:
This PR remove the dependency of form-data package

`FormData` exists built in to NodeJS now in v18+ now.
Both `got` and `node-fetch` support encoding a spec'ed `FormData` themself now, and the `form-data` package is now considered as legacy by `node-fetch` and `got`s standards. `form-data` haven't been updated for a long while and it lacks many methods and don't really follow the specification.

`node-fetch` ships with a polyfilled FormData and ways to get Blob/File like objects backed up by the filesystem that is read lazily when needed (thanks to the `fetch-blob` package). The native FormData provided by NodeJS themself do also support Blob / File like objects and dosen't need to be strictly the same instances of `globalThis.Blob` so `node-fetch` can now also export the native `FormData` when it's available. 

---

This might be a little bit of topic but i think you should strive towards eventually being able to replace `node-fetch` and `got` with built in features provided by NodeJS themself to have fewer dependencies. a long term goal and a granular replacement would be to: 
- [ ] Switch out `got` for `node-fetch`
- [ ] Switch out `node-fetch` for `Undici` when dropping support for NodeJS v16.8
- [ ] Switch out `Undici` for built in Fetch  when dropping support for NodeJS v18
- [ ] Switch out `fetch-blob/from.js` for  native fn when NodeJS adds a feature to get blob/files backed up by the file system

### DIFF
--- a/lib/plugin/gitlab/GitLab.js
+++ b/lib/plugin/gitlab/GitLab.js
@@ -2,7 +2,7 @@ import fs from 'node:fs';
 import path from 'node:path';
 import got from 'got';
 import { globby } from 'globby';
-import FormData from 'form-data';
+import { FormData, fileFromSync } from 'node-fetch';
 import allSettled from 'promise.allsettled';
 import _ from 'lodash';
 import Release from '../GitRelease.js';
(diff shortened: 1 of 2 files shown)


webpro:
Thanks @jimmywarting,  it's in [v15.6.0](https://github.com/release-it/release-it/releases/tag/15.6.0).

Totally agree with your suggestions, will keep it in mind.

PR #963, opened 2022-12-21
feat(git): added option to use --exclude option on git describe

drmartini7:
Hello,

There is a small problem in which it isn't possible to exclude certain tags when generating new releases.

For example this image:
Screen Shot 2022-12-21 at 15 14 16

Latest commit has tag `15.5.2-rc.0`, previous one has tag `15.5.1`. I ran the `git describe --tags --abbrev=0` as is used in [GitBase](https://github.com/release-it/release-it/compare/master...drmartini7:release-it:master?expand=1#diff-3a1c94688e9b16bb758da2b80710601a1966f46fd660ce47c8512a69e66d8715R97), only changing the match argument:

 - No match, obviously just returns latest tag;
 - `--match='[!-]*'`, as suggested [here](https://github.com/release-it/release-it/compare/master...drmartini7:release-it:master?expand=1#diff-1397ee01a8fd0f298ff6bb0e96747733fde00ced7fd635441e5b562228c62ff1L58), it just returns the latest tag. Why it doesn't work eludes me, as according to [glob(7) documentation](https://man7.org/linux/man-pages/man7/glob.7.html), it should work.
 - `--match='[0-9]*.[0-9]*.[0-9]*'`: this one is suggested [here](https://github.com/release-it/release-it/compare/master...drmartini7:release-it:master?expand=1#diff-fbc0ca3ee709843fbdd268266c349de4f32234fbe35d11084da4c1519c44ec4eL49), but it doesn't work since it uses the `*`, which allows any character, so everything fits.
 - `--exclude='*[-]*'`: this is the way git expects you to exclude a tag from the search as far as I could gather, so my PR aims to include this option on release-it.

There is another problem (which I guess has to do with this PR) that I could see on the tests and documentation, which I guess we could discuss here or I could open an issue to discuss it, but basically, many places on release-it use or suggest a `--match` which is something like `[0-9].[0-9].[0-9]`, like [this test](https://github.com/release-it/release-it/blob/master/test/git.init.js#L197) for example. But this way only works for up to `9.9.9`, never any double digits.

Some places use `*` to allow for more digits, but this has the side effect of allowing commits from pre-releases, which might not be what the user wants, like [here](https://github.com/release-it/release-it/compare/master...drmartini7:release-it:master?expand=1#diff-1397ee01a8fd0f298ff6bb0e96747733fde00ced7fd635441e5b562228c62ff1R56) for example.

Anyway, this isn't strictly to do with my PR, so it can be put aside or discussed on an issue if you wish.

Best Regards.

### DIFF
--- a/config/release-it.json
+++ b/config/release-it.json
@@ -11,6 +11,7 @@
     "commitMessage": "Release ${version}",
     "commitArgs": [],
     "tag": true,
+    "tagExclude": null,
     "tagName": null,
     "tagMatch": null,
     "tagAnnotation": "Release ${version}",
--- a/docs/git.md
+++ b/docs/git.md
@@ -55,12 +55,17 @@ that this represents a glob (not a regex):
 
 Example: `git.tagMatch: "[0-9]*\.[0-9]*\.[0-9]*"`
 
-Or only `"[!-]*"`, as this would match everything that excludes a hyphen, which is normally used excusively in
-pre-releaseses.
-
 This could also be useful when using a plugin to determine the next tag:
 
 Example: `git.tagMatch: "[0-9][0-9].[0-1][0-9].[0-9]*"`
+
+## Tag Exclude
+
+Use `git.tagExclude` to override the normal behavior to find the latest tag. For example whendoing a major release and 
+you want to exclude any sort of pre-releases, use `*[-]*`, as this would exclude everything with a hyphen, which is 
+normally used exclusively in pre-releases.
+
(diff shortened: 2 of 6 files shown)


webpro:
Thanks, @drmartini7 - as I think this is an improvement I went ahead and merged your PR, it's in [v15.6.0](https://github.com/release-it/release-it/releases/tag/15.6.0). When I have more time I'll dive into the issues you mention again.

PR #966, opened 2022-12-27
Fix `npm.isCollaborator()` on npm v9

indutny:
With v9.0.0 release [npm][] has changed the CLI commands for package access. While previously it was `npm access ls-collaborators` and the output was a JSON object, now it is `npm access list collaborators` and the output is plain text unless `--json` flag is provided.

This commit adds support for npm v9.0.0 while keeping support for older npm versions.

[npm]: https://github.com/npm/cli/commit/9c32c6c8d6fc5bdfd6af685731fe26920d7e5446

### DIFF
--- a/lib/plugin/npm/npm.js
+++ b/lib/plugin/npm/npm.js
@@ -132,34 +132,43 @@ class npm extends Plugin {
     );
   }
 
-  isCollaborator() {
+  async isCollaborator() {
     const registry = this.getRegistry();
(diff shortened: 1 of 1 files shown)


indutny:
Wow, that was quick. Thanks!

webpro:
Thanks for the fix! It's in [v15.6.0](https://github.com/release-it/release-it/releases/tag/15.6.0)

indutny:
Wow, this is awesome!

PR #976, opened 2023-01-20
Add yarn installation methods

baparham:
Just thought I'd add this here in case others use yarn. If `release-it` is anti-yarn or anything, feel free to just close :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,12 +40,18 @@ way to install release-it uses npm and adds some minimal configuration to get st
 
 ```bash
 npm init release-it
+# or with yarn 1.x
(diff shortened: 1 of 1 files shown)


webpro:
I'm not against yarn, but release-it uses npm under the hood (mainly `npm publish`) and while it works fine in most regular cases, it somtimes leads to issues where people don't realize this. That's why I'm not promoting it. Perhaps this is worth an explanation in the readme, but this particular PR I'm going to close :)

baparham:
Thanks, makes sense and I appreciate the explanation

PR #982, opened 2023-02-26
feat: configurable path in git rev-list

b12k:
Allows providing commits path when `requreCommits: true`.

Makes release-it configuration aligned with `@release-it/conventional-changelog` plugin:

```javascript
const releaseItConfig = {
  plugins: {
    '@release-it/conventional-changelog': {
      preset: 'conventionalcommits',
      path: '.',
      gitRawCommitsOpts: {
        path: '.',
      },
    },
  },
  git: {
    requireCommits: true,
    commitsPath: '.',
    [...]
  },
  [...]
};
```

This configuration is handy when release-it is executed in npm workspace where only related commits should be taken into account.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
+.idea
+.vscode
+*.log
+
 node_modules
 package-lock.json
 tmp
--- a/config/release-it.json
+++ b/config/release-it.json
@@ -6,6 +6,7 @@
     "requireBranch": false,
     "requireUpstream": true,
     "requireCommits": false,
+    "commitsPath": "",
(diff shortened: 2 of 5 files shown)


webpro:
Nice @b12k! I like it, happy to merge. Any chance you could add a test for this?

b12k:
@webpro 
I would like to write a test (even tried) for it but found it complicated to understand how testing works... =(

BTW I have same failing test on freshly cloned repo on win 11 (not only ubuntu). Maybe git output changed with a newer versions.

b12k:
@webpro tests added

PR #984, opened 2023-02-28
feat: make exit 0 possible when commits required but missing

b12k:
This PR introduces a possibility to exit with 0 code (success) when `requireCommits: true` and no commits available.

Why this is handy:

Alongside with this PR https://github.com/release-it/release-it/pull/982 it will make release-it a monorepo best friend.

Some times it is OK not to have commits related to a certain workspace. Exit with error may stop the pipeline execution, depending on CI configuration.

This particular feature is inspired by monorepo managed by turborepo where every workspace has a `"release": "release-it"` npm script in `package.json`.

When turbo release pipeline is executed failing releases are:
- terminating pipeline execution
- preventing execution results to be cached

I my case, right now, a custom script, which checks if there commits that touch a specific workspace, prevents an execution of release-it (not to exit with non zero code).

Its kinda specific use case but I think it can be a great addition to release-it.

### DIFF
--- a/bin/release-it.js
+++ b/bin/release-it.js
@@ -3,7 +3,7 @@
 import updater from 'update-notifier';
 import parseArgs from 'yargs-parser';
 import release from '../lib/cli.js';
-import { readJSON, ErrorExitSuccess } from '../lib/util.js';
+import { readJSON } from '../lib/util.js';
 
 const pkg = readJSON(new URL('../package.json', import.meta.url));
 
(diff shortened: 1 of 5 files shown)


webpro:
Nice! Love this PR, that `process.exit(code)` might come in handy more often. Will merge when tests pass.

webpro:
Apparently I missed a line when fixing conflicts here in the GitHub UI, merged anyway to fix locally and push into master.

PR #1001, opened 2023-04-18
lodash's `isPlainObject` method in plain js

PunitSoniME:


webpro:
Missing a description. What's the benefit of doing this?

PunitSoniME:
Getting rid of lodash dependencies

webpro:
- There are a lot more imports from `lodash` in the codebase
- Lodash is tried and trusted, why would this project be better off with this unknown implementation?
- Performance in general is not _really_ an issue for release-it in general.
- With this, the project has more code to maintain.

A more complete pull request with `lodash` replaced by something could be interesting, but Lodash is already just one dependency so not sure how to benefit here.

PR #1004, opened 2023-05-02
Skip version release if plugin `getIncrementedVersion` return null

cakeinpanic:
Hi! 
Love this package, using it in our CI infrastructure!

Added a small change to support case when plugin returns null in `getIncrementedVersion` (this is possible, [example one](https://github.com/release-it/conventional-changelog/blob/master/index.js#L58), [example two](https://github.com/release-it/conventional-changelog/blob/master/index.js#LL139C48-L139C69))


cakeinpanic:
Hi, @webpro, please take a look :)

Issue #997, opened 2023-04-05
Even if private registry is used, log refers to npmjs.com


cakeinpanic:
Hi, first of all thank you for the great package, I love it!

Noticed this: I am using private registry(it's set in `.npmrc`), and release works smooth, however log says this
```
🔗 https://www.npmjs.com/package/yotpo-dependabot-test
```
which is misleading

As an easy fix I suggest to support one more config property `npm.privateRegistry : boolean` and do not post a link to a package if it's set to true. WDYT?

webpro:
Currently release-it reads only `publishConfig.registry` and falls back to the default registry.

Looks like we can use `npm config get registry` to read that value if it's in `.npmrc`, so that could be another way to approach this without needing more config. WDYT?

Any chance you're up for a PR? Otherwise I'll pick it up within a week or so.

Thanks for the kind words :)

PR #1009, opened 2023-05-24
fix: use deep merge strategy for context and options in Plugin

xllily:
Initially, I attempted to modify the "source" order in `Object.assign`, but upon further consideration, I realized that utilizing a deep merge strategy would be a safer approach.

### DIFF
--- a/lib/plugin/Plugin.js
+++ b/lib/plugin/Plugin.js
@@ -1,6 +1,5 @@
 import { debug } from 'node:util';
 import _ from 'lodash';
-import { deepMerge } from '../util.js';
 
 class Plugin {
   static isEnabled() {
@@ -41,7 +40,7 @@ class Plugin {
   afterRelease() {}
 
   getContext(path) {
-    const context = deepMerge({}, this.options, this.context)
+    const context = _.merge({}, this.options, this.context);
     return path ? _.get(context, path) : context;
   }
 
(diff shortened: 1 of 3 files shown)


webpro:
Thanks @xllily! Could you please give an example of what this change will fix? And given that Lodash is already a dependency, why did you choose to add a custom `deepMerge`?

xllily:
> Thanks @xllily! Could you please give an example of what this change will fix? And given that Lodash is already a dependency, why did you choose to add a custom `deepMerge`?

Thank you @webpro  for your feedback! I apologize for not providing an example initially.

Regarding the issue with the log output in the terminal after a successful release, let me explain further. The example `.release-it.json` configuration you provided includes a custom registry URL (`https://{{my-custom-host}}/`) for npm publishing. However, despite this configuration, the log in the terminal still displays the default npm registry URL (`https://www.npmjs.com/package/{{my-package}}`).

I understand that this behavior is unexpected, and it aligns with the issue mentioned in #997. The purpose of my pull request was to address this issue and ensure that the log output in the terminal reflects the custom registry URL specified in the configuration.

As for the inclusion of the custom `deepMerge` implementation, I have realized that it may not be necessary in this case. I understand that Lodash is already a dependency, and its `merge` function can effectively handle the merging requirements. I have updated the pull request accordingly to reflect this realization.

Once again, thank you for your valuable feedback and guidance. I appreciate your assistance in improving the pull request, and I apologize for any confusion caused by my initial approach. Please let me know if there are any further concerns or suggestions.

webpro:
Thanks @xllily! This is great, happy to merge.

webpro:
:rocket: _This pull request is included in v15.10.4. See [Release 15.10.4](https://github.com/release-it/release-it/releases/tag/15.10.4) for release notes._

PR #1020, opened 2023-07-04
Use all refs to determine latest version

### LINKED ISSUE #1016: Most recent tag is taken from only the current branch
To calculate the next version from git tags, release-it uses `git describe` to find the latest tag:
https://github.com/release-it/release-it/blob/14a73d977a30a817de5896a65f58a4069d8d7c88/lib/plugin/GitBase.js#L93-L101

The issue is that `git describe` can only find tags on the same branch, because it finds the most recent tag reachable from a commit (e.g. HEAD of the checked-out branch). This is not necessarily the same as the latest tag across all branches.

For example, consider a git flow approach where you want to tag builds from `main` as normal/full versions such as `v1.0.0`, but builds from `develop` with a suffix such as `-dev.X`:
1. `main` and `develop` are both at `v1.0.0`
2. Merging a feature branch to `develop` incrementing patch & modifier tags it as `v1.0.1-dev.0`
3. Merging `develop` to `main` incrementing patch tags it `v1.0.1`, which is great
4. Now you have `v1.0.1` pointing to `main` and `v1.0.1-dev.0` pointing to `develop`
5. Merging a new feature to `develop` you want it to be `v1.0.2-dev.0`, which is what incrementing from `v1.0.1` would give, but instead `git describe` finds `v1.0.1-dev.0` as the latest tag and so the increment results in `v1.0.1-dev.1`

Even though you are merging to `develop`, the real latest tag for the whole project is on `main`. What do you recommend here? `git tag --list` might be a solution, it will return all tags, but they would need sorting properly. Unsure if this would actually be undesirable, and some people *do* want the current latest tag from same branch only via `git describe` 🤔

imgrant:
This PR adds a configuration parameter, `git.getLatestTagFromAllRefs`, which will use `git for-each-ref` instead of `git describe` to find the latest version from git tags.

`git describe` finds the most recent tag that is reachable from the current commit; newer tags could exist, pointing to commits on different branches, which `git describe` would not find. `git for-each-ref` is used to list all tags, matching and sorting appropriately, to find the most recent semantically-ordered tag.

An illustration is provided in the [documentation](docs/git.md). The default behaviour is unchanged; the change is not breaking.

Resolves #1016


webpro:
Thanks @imgrant, much appreciated! I also like the `getLatestTagFromAllRefs` naming. Any chance you could add some coverage with a test? Just let me know, I'd be happy to do it myself.

imgrant:
> Thanks @imgrant, much appreciated! I also like the `getLatestTagFromAllRefs` naming. Any chance you could add some coverage with a test? Just let me know, I'd be happy to do it myself.

Sure! [3fbb6a4](https://github.com/release-it/release-it/pull/1020/commits/3fbb6a48c04729cb5a4b1672cd6e3afe1de7d682) is it — I wasn't too sure about all the testing stuff tbh, feel free to refactor. But, it passes with `git.getLatestTagFromAllRefs: true` and fails with `git.getLatestTagFromAllRefs: false` (or default).

webpro:
Yesterday there were some issues with transitive dependencies in older Node.js versions, eventually decided to do a major release, add a lockfile and drop support for Node.js v14. Once everything seems stable out there I'll merge this!

webpro:
:rocket: _This pull request is included in v16.1.0. See [Release 16.1.0](https://github.com/release-it/release-it/releases/tag/16.1.0) for release notes._

webpro:
Thanks again @imgrant, this is such a great addition to have!

imgrant:
No worries, thanks for making the process to integrate it seamless!

PR #1055, opened 2023-10-31
feat: add Typescript type for config (closes #1052)

### LINKED ISSUE #1052: TypeScript Config file
Just wondering, is there any reason `.ts` isn't included in the list of searches for config files?

Would be nice to make a typed config file.

I'd be happy to submit a PR too :)

JohnCampionJr:
I only added the Config; typing the whole project would only be useful for plugin authors and would be a much larger project.

Sample ts config:
```ts
import type { Config } from 'release-it'

export default {
  plugins: {},
  hooks: {},
  git: {
    commit: true,
    tag: true,
    push: true,
  },
  github: {
    release: true,
  },
  npm: {
    publish: true,
  },
} satisfies Config
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ This document lists breaking changes for each major release.
 See the GitHub Releases page for detailed changelogs:
 [https://github.com/release-it/release-it/releases](https://github.com/release-it/release-it/releases)
 
+## v17 (2023-11-11)
+
+- Removed support for Node.js v16.
+
 ## v16 (2023-07-05)
 
 - Removed support for Node.js v14.
--- a/README.md
+++ b/README.md
@@ -337,8 +337,9 @@ release-it programmatically][48] for example code.
 
 ## Legacy Node.js
 
-The latest major version is v16, supporting Node.js 16 and up (as Node.js v14 is EOL). Use release-it v15 for
-environments running Node.js v14. Also see [CHANGELOG.md][70].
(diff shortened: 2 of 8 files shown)


webpro:
Thanks @JohnCampionJr! I'm happy to merge it, but is there a way to automate the generation of the `types/config.d.ts` file?

JohnCampionJr:
There probably is but I’m not sure how without moving everything to
typescript. This is more of a stop-gap.

ghiscoding:
Would it be possible to also add brief instructions on 1 of the README so it would be easier for discovery. Thanks

JohnCampionJr:
I've added a bit to the config docs

JohnCampionJr:
Looks like you just did that; I don't see any unrelated changes any more either.

webpro:
And how about adding it to [Definitely Typed](https://github.com/DefinitelyTyped/DefinitelyTyped)?

Inspired by https://github.com/JoshuaKGoldberg/release-it-action/issues/29

northword:
Also, I think that the command line arguments (constructor arguments) could include type declarations as well, since they can affect release-it as well. 

example: 

https://github.com/northword/zotero-plugin-scaffold/blob/03ca4f76205e6fa4201ec1a2c14e782c1bb216bd/types/release-it.d.ts#L92-L100

webpro:
Thanks @JohnCampionJr, much appreciated. Sorry it took so long.

webpro:

Issue #1058, opened 2023-11-04
Still pushes a tag when HEAD 


cakeinpanic:
Hi, ran into this situation:

when I'm running `release-it` on a commit that's _behind current remote HEAD_, process fails and I'm getting this(completely legit!) error
```
ERROR To https://github.com/cakeinpanic/test-repo
 * [new tag]         pacakge@4.0.1 -> pacakge@4.0.1
 ! [rejected]        master -> master (non-fast-forward)
error: failed to push some refs to 'https://github.com/YotpoLtd/katya-test'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. If you want to integrate the remote changes,
hint: use 'git pull' before pushing again.
```
But the problem is that tag is being pushed and not being deleted afterwards, which leads to a situation when I have a legit tag on a commit does not belong to any branch on this repository.

### Suggested fix
I suggest either
1. Push commits first and then if there was no error, push new tag
2. Revert pushed tag if push of commits failed

I'd be glad to provide such a PR on it, want to hear your opinion first

ghiscoding:
Hello, I just followed the link to see what the issues were fix in latest release and I think this issue was by PR #1059 and should be closed, am I right? 

Side note adding the issue number in a PR will automatically close the issue when the PR is merged, something like this `fix: rollback tag, fixes #1058` would have closed the issue when merge. 

Cheers

webpro:
You're right, thanks Ghislain! That'll do it automagically next time :)

PR #1059, opened 2023-11-04
Fix/rollback tag

cakeinpanic:
Hi! 
Fixing issue described in https://github.com/release-it/release-it/issues/1058

I tried to use standard rollback method, but because pushing to remote is async, it does not work on the `exit` hook

If it's crucial to have all rollback in one method, there are few options I have found 
1. Don't call `exit` explicitly and then get ability to use `beforeExit` hook 
2. Export rollback method and explicitly call it before exiting
3. Give a try to https://www.npmjs.com/package/async-kit (did not experiment with it yet)

WDYT?


cakeinpanic:
@webpro please take a look, this thing is blocking us from using the tool :(

webpro:
Thanks @cakeinpanic! It's an interesting feature, complete with a good test, let's try it out!

cakeinpanic:
@webpro hi! altogether with this fix you have released node 16 deprecation altogether with dependency packages updates.

we are using node 14.20 and can not upgrade to node 17 at the moment

and started getting 
```
> release-it --ci
node_modules/release-it/node_modules/update-notifier/update-notifier.js:148
                options.boxenOptions ??= {
```

is there a way to release also next minor on v16 with this fix only?

webpro:
:rocket: _This pull request is included in v16.3.0. See [Release 16.3.0](https://github.com/release-it/release-it/releases/tag/16.3.0) for release notes._

webpro:
Alright then, although release-it v16 also required Node.js v16

webpro:
I don't fancy going to back to supporting Node.js v14 or earlier, but I'll happily accept fixes against the v16 branch.

cakeinpanic:
@webpro thank you for fast response! release v 16 actually works fine with node 14 🫠 
but with a warning 
```
UnhandledPromiseRejectionWarning: TypeError: Intl.Segmenter is not a constructor
    at stringWidth (file:///node_modules/release-it/node_modules/ora/node_modules/string-width/index.js:27:37)
```

but i have made a mistake, we are using version 15 in our repos...

I have actually found an uuuugly workaround using hooks, since revert of published tag is not our only issue – we dont want to publish an npm package at all if we ar enot able to push, so with my hook workaround this fix is not even needed :)

webpro:
lol

Issue #1095, opened 2024-04-10
Return different message when requireCommitsFail is false


juancarlosjr97:
I have been thinking that when `git.requireCommitsFail` is set to `false` the message `ERROR There are no commits since the latest tag.` should not return `ERROR` but instead should return something else, like `INFO`.

It fails safe but still keeps the same message which I think is misleading in terms of logs.

Any thoughts?

webpro:
I think you're right 💯

Feel free to open a PR for this :)

PR #1101, opened 2024-04-25
Exit without error message when requireCommits is true but requireCommitsFail is false and no commits are available 

juancarlosjr97:
### Change

Updating the flow to when requireCommits is true but requireCommitsFail is set as false and no commits are available, instead of returning ERROR There are no commits since the latest tag. just returns There are no commits since the latest tag. and remove the ERROR string as that is misleading.

Fixes https://github.com/release-it/release-it/issues/1095

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -138,7 +138,12 @@ const runTasks = async (opts, di) => {
     };
   } catch (err) {
     const { log } = container;
-    log ? log.error(err.message || err) : console.error(err); // eslint-disable-line no-console
+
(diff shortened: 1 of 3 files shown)


webpro:
:rocket: _This pull request is included in v17.2.1. See [Release 17.2.1](https://github.com/release-it/release-it/releases/tag/17.2.1) for release notes._

webpro:
Thanks a bunch, @juancarlosjr97!

PR #1106, opened 2024-05-14
feat(packages): updates Node.js version range to include `22.x.x`

adam-sokolowski:


webpro:
I'm sorry, I saw your PR too late, but also I would want the v21 to be removed so here we are :)

Still, thanks a lot for your support!

PR #1111, opened 2024-05-22
chore: Upgrade ESLint

nzakas:
This PR upgrades ESLint to v9.0, including the creation of the new config file.

One thing to note: `eslint-plugin-import` has a peer dependency on ESLint < 9, so in order to install, you'll need to do `npm install -f` until they are able to update that dependency. I updated the `test.yml` workflow to do that.

I also ran ESLint with `--fix`, which removed some of the unneeded disable directives.

Thanks for putting this forward as a test for the config migration tool I'm working. It proved to find several bugs.


webpro:
Thanks @nzakas! I think `--legacy-peer-deps` is slightly better than `-f` so I went with that. Glad it was helpful and release-it is on ESLint v9!

webpro:

Issue #1076, opened 2024-01-21
provide json schema of config file


hezhengxu2018:
like lerna, the json schema path has added to its config file after `lerna init`.
`"$schema": "node_modules/lerna/schemas/lerna-schema.json"`
provide the json schema would improve user experience，user can write config file without doc.

ChromeGG:
Bump, also looking for this feature :)

webpro:
That would be a nice feature indeed, if someone is willing to create and submit a JSON Schema file that would be great.

There's also ideas to add types in #1055.

juancarlosjr97:
Cool idea! What is everyone's thought on having it centralized instead of part of the installation, the same as Renovate schema for example.

```json
  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
```

Sukaato:
I hope you're happy with these json schemas :)

webpro:
This has also been added to create-release-it (https://github.com/release-it/create-release-it/blob/master/index.js#L40-L43) so it comes with `npm init release-it` by default.

voxpelli:
I guess this can be closed as fixed then? :)

webpro:
Yes, thanks again @Sukaato! Very appreciated.

PR #1114, opened 2024-06-12
feat: Add json schema

Sukaato:
#1076 
Add json schema for configure release-it.json

### DIFF
--- a/schema/release.json
+++ b/schema/release.json
@@ -1,0 +1,55 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "node_modules/release-it/schema/release",
+  "title": "JSON schema for Relase-it configuration",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "hooks": {
+      "type": "object",
+      "additionalProperties": true,
+      "patternProperties": {
+        "^(before|after):((version|git|npm|github|gitlab):)?(init|bump|release)$": {
+          "if": {
+            "type": "array"
+          },
+          "then": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            }
+          },
+          "else": {
+            "type": "string"
+          }
+        }
+      }
+    },
+    "plugins": {
+      "type": "object",
+      "additionalProperties": true
+    },
+    "git": {
+      "$comment": "Can be a boolean or git config object",
+      "if": {
+        "type": "object"
+      },
+      "then": {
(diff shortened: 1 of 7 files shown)


schema/release.json
```diff
@@ -0,0 +1,55 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "node_modules/release-it/schema/release",
```
webpro, line 3: Why does this value contain `node_modules`? Is it supposed to reference itself?

schema/release.json
```diff
@@ -0,0 +1,55 @@
...
+      "type": "object",
+      "additionalProperties": true
+    },
+    "git": {
+      "$comment": "Can be a boolean or git config object",
+      "if": {
+        "type": "object"
+      },
+      "then": {
+        "$ref": "node_modules/release-it/schema/git.json"
```
webpro, line 38: Maybe this could be a relative reference to `"./git.json"`?

schema/release.json
```diff
@@ -0,0 +1,55 @@
...
+      },
+      "else": {
+        "type": "boolean",
+        "default": false
+      }
+    },
+    "npm": {
```
webpro, line 45: Every plugin can be `false` (like `"git"`)

schema/test.json
```diff
@@ -0,0 +1,6 @@
+{
+  "$schema": "file:///C:/Users/KANOMA1/Documents/projects/release-it/schema/release.json",
```
webpro, line 2: Maybe this file should be omitted entirely?

webpro (commented):
Thanks for the PR, @Sukaato. I have a few questions. Also: can this file be auto-generated somehow?

schema/release.json
Sukaato: I've just fixed all $ref path

schema/release.json
Sukaato: it's just a unique id, we can write we want, i've just change it for a better in latest commit

schema/release.json
Sukaato: fixed in last commit

schema/test.json
Sukaato: removed in last commit

webpro:

webpro:
Thanks @Sukaato! This is the recommendation:

```json
{
  "$schema": "https://unpkg.com/release-it@17/schema/release-it.json"
}
```

I'll add it to the docs soon.

PR #1120, opened 2024-06-27
chore: drop dependency on promise.allsettled

brc-dd:
Promise.allSettled has been supported since node v12 (https://node.green/#ES2020-features--Promise-allSettled) and release-it only supports `"node": "^18.18.0 || ^20.9.0 || ^22.0.0"`. We can safely drop the dependency on `promise.allsettled` ponyfill. Moreover, the `promise.allsettled` dep is pulling 4MB of unnecessary code.

Also bumps inquirer which recently dropped unnecessary dependencies. This will reduce release-it's install size by over 13%.

PS: Using both `node-fetch` and `got` seems wrong - just one will be sufficient. Moreover, they can probably be replaced by node's built-in fetch which was unflagged in node v18.0.

### DIFF
--- a/lib/plugin/gitlab/GitLab.js
+++ b/lib/plugin/gitlab/GitLab.js
@@ -3,7 +3,6 @@ import path from 'node:path';
 import got from 'got';
 import { globby } from 'globby';
 import { FormData, fileFromSync } from 'node-fetch';
-import allSettled from 'promise.allsettled';
 import _ from 'lodash';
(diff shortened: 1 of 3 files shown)


webpro:
Thanks @brc-dd, much appreciated!

webpro:
:rocket: _This pull request is included in v17.4.1. See [Release 17.4.1](https://github.com/release-it/release-it/releases/tag/17.4.1) for release notes._

PR #1123, opened 2024-07-02
fix: vulnerable packages

amarbardolia:
`release-it` has transitive dependency on `ip` package (https://github.com/indutny/node-ip) which was recently reported as vulnerable. 

Here is the description of the **dependabot** alert raised by GitHub on one of our private repos which uses `release-it` 

> The ip package through 2.0.1 for Node.js might allow SSRF because some IP addresses (such as 127.1, 01200034567, 012.1.2.3, 000:0:0000::01, and ::fFFf:127.0.0.1) are improperly categorized as globally routable via isPublic. NOTE: this issue exists because of an incomplete fix for https://github.com/advisories/GHSA-78xj-cgh5-2h22.

`pac-resolver` and `socks` packages were upgraded to get rid of ip package

https://github.com/TooTallNate/proxy-agents/pull/281
https://github.com/JoshGlazebrook/socks/pull/94

Alongside also upgraded vulnerable `braces` package to 3.0.3

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1982,11 +1982,11 @@
       }
     },
     "node_modules/braces": {
-      "version": "3.0.2",
-      "resolved": "https://registry.npmjs.org/braces/-/braces-3.0.2.tgz",
-      "integrity": "sha512-b8um+L1RzM3WDSzvhm6gIz1yfTbBt6YTlcEKAvsmqCZZFw46z626lVj9j1yEPW33H5H+lBQpZMP1k8l+78Ha0A==",
+      "version": "3.0.3",
+      "resolved": "https://registry.npmjs.org/braces/-/braces-3.0.3.tgz",
+      "integrity": "sha512-yQbXgO/OSZVD2IsiLlro+7Hf6Q18EJrKSEsdoMzKePKXct3gvD8oLcOQdIzGupr5Fj+EDe8gO/lxc1BzfMpxvA==",
       "dependencies": {
-        "fill-range": "^7.0.1"
+        "fill-range": "^7.1.1"
       },
       "engines": {
         "node": ">=8"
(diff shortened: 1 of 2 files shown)


webpro:
Thanks @amarbardolia! Shouldn't there be a change in `package.json` to make the override more explicit?

amarbardolia:
@webpro Thanks for the review. I can't think of a reason to update `package.json` file. And FYI the changes I added were the outcome of `npm audit fix`. Let me know if you still think there should be a change in `package.json` file with a snippet and would be happy to make a change.

webpro:
I mean overrides: https://docs.npmjs.com/cli/configuring-npm/package-json#overrides

amarbardolia:
@webpro you mean we should add overrides in our repo's package.json instead of updating `release-it`?

webpro:
No, in the `package.json` of release-it itself, so the package manager knows about it. The lockfile isn't published!

amarbardolia:
@webpro Thanks. Added `overrides` for vulnerable packages.

webpro:
> `pac-resolver` and `socks` packages were upgraded to get rid of ip package
> 
> [TooTallNate/proxy-agents#281](https://github.com/TooTallNate/proxy-agents/pull/281) [JoshGlazebrook/socks#94](https://github.com/JoshGlazebrook/socks/pull/94)

Thanks!
 
> Alongside also upgraded vulnerable `braces` package to 3.0.3

Why did you upgrade `braces` and why is it in `overrides`? It doesn't seem to be using `ip`.

We could upgrade it, but seems to me it's a different thing that's not supposed to go into `overrides` (nor in this PR).

webpro:
Sorry, we could pick that up in a separate PR if you still want?

webpro:
Can't we also just upgrade proxy-agent now, instead of using the overrides? See https://github.com/TooTallNate/proxy-agents/commits/main/

webpro:
I'll update some dependencies now and publish, then we can do another scan?

webpro:
Unfortunately the container package proxy-agent wasn't updated upstream.

webpro:
Thanks @amarbardolia!

webpro:
:rocket: _This pull request is included in v17.4.2. See [Release 17.4.2](https://github.com/release-it/release-it/releases/tag/17.4.2) for release notes._

amarbardolia:
> Unfortunately the container package proxy-agent wasn't updated upstream.

Yup. That's why couldn't upgrade it. Thanks @webpro  for merge.

PR #1124, opened 2024-07-04
fix(types): requireBranch is either a string, false, or undefined

timmywil:
Hit a type error when using `requireBranch`. The correct type seems to be a `string`, `undefined`, or `false`, which is apparently the default.


webpro:
Thanks @timmywil!

webpro:
:rocket: _This pull request is included in v17.4.2. See [Release 17.4.2](https://github.com/release-it/release-it/releases/tag/17.4.2) for release notes._

Issue #1125, opened 2024-07-04
Allow insecure connections to self-hosted Gitlab instances


vonBrax:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `release-it@17.4.1` for the project I'm working on.

The problem we have is that we host our own private Gitlab instance deployed to our internal network using a self signed certificate.

I understand there is already an option to pass our certificate to `release-it`, but it is not very ergonomic since we need to keep passing the certificate around to everyone in the project before they can create a release.

The [got](https://github.com/sindresorhus/got) library used internally by `release-it` already provides an [option](https://github.com/sindresorhus/got/blob/ae04c0e36cf3f5b2e356df7d48a5b19988f935a2/source/core/options.ts#L1960) to allow insecure https connection.

I was wondering if it would be possible to extend the Gitlab options to also accept a `secure` flag and pass it down to `got`?

I'll be more than happy to open a PR if you accept this suggestion.

Here is the diff that solved our problem:

```diff
diff --git a/node_modules/release-it/lib/plugin/gitlab/GitLab.js b/node_modules/release-it/lib/plugin/gitlab/GitLab.js
index 60dc8ac..fd20c81 100644
--- a/node_modules/release-it/lib/plugin/gitlab/GitLab.js
+++ b/node_modules/release-it/lib/plugin/gitlab/GitLab.js
@@ -169,7 +169,13 @@ class GitLab extends Release {
   }
 
   async request(endpoint, options) {
-    const { baseUrl } = this.getContext();
+    const { baseUrl, secure } = this.getContext();
+    
+    if (secure === false) {
+      options.https = options.https || {};
+      options.https.rejectUnauthorized = false;
+    }
+
     this.debug(Object.assign({ url: `${baseUrl}/${endpoint}` }, options));
     const method = (options.method || 'POST').toLowerCase();
     const response = await this.client[method](endpoint, options);
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

webpro:
> I was wondering if it would be possible to extend the Gitlab options to also accept a `secure` flag and pass it down to `got`?
> 
> I'll be more than happy to open a PR if you accept this suggestion.

Sounds good! Looks all reasonable and sound :)

PR #1129, opened 2024-07-09
feat: add secure flag to gitlab config

vonBrax:
This PR adds a new flag to the gitlab config options: `secure`.

If the value of this flag is **not** `undefined`, it will be forward to https request library (`got`), which in turn also forwards it to node's `https.request` method as the `rejectUnauthorized` flag.

The `rejectUnauthorized` flag is `true` by default. When its value is `false`, it allows node to make https requests to servers without verifying the server certificate against the list of trusted CA's.

The goal of this PR is to reduce the effort to create gitlab releases for teams working with self-hosted private gitlab instances, skipping the need of having to pass CA certificates around.

Fix: #1125

### DIFF
--- a/lib/plugin/gitlab/GitLab.js
+++ b/lib/plugin/gitlab/GitLab.js
@@ -23,7 +23,9 @@ class GitLab extends Release {
       certificateAuthority: certificateAuthorityFile ? fs.readFileSync(certificateAuthorityFile) : undefined,
       rejectUnauthorized: typeof secure === 'boolean' ? secure : undefined
     };
-    const https = _.merge({}, httpsOptions);
+
+    // Remove keys with undefined values
+    const https = _.pickBy(httpsOptions, value => value !== undefined);
 
     this.certificateAuthorityOption = _.isEmpty(https) ? {} : { https };
   }
--- a/lib/plugin/gitlab/GitLab.js
+++ b/lib/plugin/gitlab/GitLab.js
@@ -17,10 +17,15 @@ class GitLab extends Release {
     super(...args);
     this.registerPrompts(prompts);
     this.assets = [];
-    const { certificateAuthorityFile } = this.options;
-    this.certificateAuthorityOption = certificateAuthorityFile
-      ? { https: { certificateAuthority: fs.readFileSync(certificateAuthorityFile) } }
-      : {};
+    const { certificateAuthorityFile, secure } = this.options;
+
+    const httpsOptions = {
+      certificateAuthority: certificateAuthorityFile ? fs.readFileSync(certificateAuthorityFile) : undefined,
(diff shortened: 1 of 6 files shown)


lib/plugin/gitlab/GitLab.js
```diff
@@ -17,10 +17,15 @@ class GitLab extends Release {
...
-    this.certificateAuthorityOption = certificateAuthorityFile
-      ? { https: { certificateAuthority: fs.readFileSync(certificateAuthorityFile) } }
-      : {};
+    const { certificateAuthorityFile, secure } = this.options;
+
+    const httpsOptions = {
+      certificateAuthority: certificateAuthorityFile ? fs.readFileSync(certificateAuthorityFile) : undefined,
+      rejectUnauthorized: typeof secure === 'boolean' ? secure : undefined
+    };
+    const https = _.merge({}, httpsOptions);
```
webpro, line 26: Looks like we don't need this intermediate step?

webpro (commented):
Looks good! One nitpick, but also a request: any chance you could add some test coverage? There's a few for the GitLab plugin already.

lib/plugin/gitlab/GitLab.js
vonBrax: That's correct, we don't really need it. My idea here was to use lodash merge because it remove keys with undefined values, so we wouldn't end up setting `https` to an empty object, which was what the code was doing before. But I can remove this line.

Regarding testing, I tried to find a test using CA certificates, but couldn't find one, so I skipped it. Or do you mean just a unit test to check if the options passed to the Gitlab plugin are correct?

webpro: > That's correct, we don't really need it. My idea here was to use lodash merge because it remove keys with undefined values, so we wouldn't end up setting `https` to an empty object, which was what the code was doing before. But I can remove this line.

Thanks, `merge` is not descriptive towards that end, but let's just keep it then.

> Regarding testing, I tried to find a test using CA certificates, but couldn't find one, so I skipped it. Or do you mean just a unit test to check if the options passed to the Gitlab plugin are correct?

Just the latter!

test/gitlab.js
vonBrax, line 1: @webpro 

I decided to do multiple checks inside the same test, because they are fairly simple, but I can also split them into multiple tests if you prefer.

webpro: Cool! Or use blocks: https://www.webpro.nl/scraps/javascript-block-statement :)

vonBrax: Done 🫡

webpro:
:rocket: _This pull request is included in v17.6.0. See [Release 17.6.0](https://github.com/release-it/release-it/releases/tag/17.6.0) for release notes._

webpro:
Thanks @vonBrax!

PR #1132, opened 2024-07-11
feat: support proxy env variables in gitlab plugin

SebastianGoeb:


lib/plugin/github/GitHub.js
```diff
@@ -182,9 +182,7 @@ class GitHub extends Release {
       }
     };
 
-    if (proxy) {
-      options.request.agent = new ProxyAgent(proxy);
-    }
+    options.request.agent = new ProxyAgent(proxy);
```
webpro, line 185: Why this change? Same for GitLab plugin.

I guess what we want here is a way to instantiate `ProxyAgent()` without arguments? Rather than always instantiate it using `null` by default.

How about adding `true` as an option value for `github.proxy` and `gitlab.proxy`?

webpro (commented):
Thanks @SebastianGoeb, thanks the PR. One comment, though.

lib/plugin/github/GitHub.js
SebastianGoeb: `ProxyAgent` can be instantiated without arguments, in which case `http_proxy` etc will be used. So by always instantiating a `new ProxyAgent()` we should always have support for proxy env vars, if defined, and possibly some specific overrides. And in GitHub's case, if `options.proxy` is unset, then the resulting call should become `new ProxyAgent(undefined)`, which is just the same as `new ProxyAgent()`.

I was hoping to achieve this behavior:
| http_proxy vars set | options.proxy set | result |
| --- | --- | --- |
| no | no | no proxy used |
| yes | no | proxy from env used |
| no | yes | proxy from option.proxy used |
| yes | yes | proxy from env + overrides from option.proxy used |

I would really like to test that all the combinations work right though...

webpro: I think it should default to not use/instantiate `ProxyAgent` at all.

SebastianGoeb: I'm not quite I understand. The whole premise of this PR was to enable the use of `http_proxy` vars by default, meaning a `new ProxyAgent()` would have to be instantiated unconditionally in all cases.

Do you mean, you would rather keep it like this, which is to say "only use proxy if options.proxy blocks exists, no matter if empty or not, but then do pull in env vars"
```
# for both gitlab and github options
if (this.options.proxy) {
  # also pulls in env vars, even if this.options.proxy == {}
  options.request.agent = new ProxyAgent(this.options.proxy); 
}
```

Or do you mean something more like an explicit "I want to use system env vars" option, sorta like this:
```
# for both gitlab and github options
if (this.options.proxy?.useEnvVars) {
  options.request.agent = new ProxyAgent(); 
} else if (this.options.proxy) {
  options.request.agent = new ProxyAgent(this.options.proxy); 
}
```

webpro: > Do you mean, you would rather keep it like this, which is to say "only use proxy if options.proxy blocks exists, no matter if empty or not, but then do pull in env vars"

Yes, this. That's why I mentioned this:

> How about adding `true` as an option value for `github.proxy` and `gitlab.proxy`?

I could imagine it uses config options if set, and falls back to env vars - i.e. always tries to read env vars if `github.proxy` is empty object or `true`. Which I think also means the least of surprises/breaking changes for existing users of release-it.

webpro: If there's still interest, perhaps it's best to hook this one into #1133

webpro:
Closing this due to inactivity. But feel free to open a new PR, happy to look into it. Probably good to focus on GitLab or GitHub separately, though.

PR #1150, opened 2024-09-14
feat(github): expose github make_latest flag

sirchnik:
In a project we create non-latest support release regularly. It is annoying to reset the latest release everytime.
Unfortunately release-it does not expose the make_latest flag from  the [GitHub API](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release).

This PR exposes this option.

Something to think about: Should the `make_latest: "legacy"` option even be exposed or should it be renamed?

I tested this here: https://github.com/sirchnik/test-npm

### DIFF
--- a/.release-it.json
+++ b/.release-it.json
@@ -1,4 +1,5 @@
 {
+  "$schema": "./schema/release-it.json",
   "hooks": {
     "after:init": ["npm run lint", "npm run knip", "npm test"]
   },
--- a/docs/github-releases.md
+++ b/docs/github-releases.md
@@ -166,6 +166,11 @@ Example command to add assets and explicitly toggle the draft status to "publish
 release-it --no-increment --no-git --github.release --github.update --github.assets=*.zip --no-github.draft
 ```
 
+## Project Non-Latest Release
+
+To do a release that isn't the latest release on your GitHub project e.g for support releases, you can set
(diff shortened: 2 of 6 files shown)


webpro:
Thanks @sirchnik! Would be happy to merge this, but I have two questions:

- Why not just take the same value options for `make_latest` as GitHub API? Then for anything they support, we don't need to change much here (other than types).
- Any chance you could add something to a (existing) test?

sirchnik:
For 1.: 
* Do you mean the values for makeLatest should be `'true' | 'false' | 'legacy'` or should I leave it as `true | false | 'legacy'`?
* Or should the name of the option be the same as in github so `make_latest`?
For 2. I'll try later to figure out what the tests do and integrate this change.

sirchnik:
I couldn't run the test on my machine (arch node 22) only in a devcontainer. I don't think for a user passable option a test is needed.

sirchnik:
@webpro Do I need to add a test for this?

webpro:
Thanks @sirchnik! Also for the other small improvements. Perfect.

webpro:
:rocket: _This pull request is included in v17.7.0. See [Release 17.7.0](https://github.com/release-it/release-it/releases/tag/17.7.0) for release notes._

PR #1153, opened 2024-09-22
Fix uploaded assets URL for Gitlab version >= 17.2

jeffesquivels:
Gitlab started to use a [new URL format](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/150939) for uploaded assets in version 17.1 and since version 17.2 all new uploaded assets are [no longer accessible via the old URL format](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/156939), which means that uploaded assets are no longer accessible from the links in the release.

This PR introduces a new flag for Gitlab which allows the user to choose which URL format release-it should use when publishing the uploaded assets in a new release.

### DIFF
--- a/schema/gitlab.json
+++ b/schema/gitlab.json
@@ -28,7 +28,7 @@
       "type": "boolean",
       "default": false
     },
-    "milesstones": {
+    "milestones": {
       "type": "array",
       "items": {
         "type": "string"
--- a/schema/gitlab.json
+++ b/schema/gitlab.json
@@ -52,6 +52,10 @@
     "assets": {
       "default": null
     },
+    "useIdsForUrls": {
+      "type": "boolean",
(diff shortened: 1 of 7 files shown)


schema/gitlab.json
```diff
@@ -28,7 +28,7 @@
       "type": "boolean",
       "default": false
     },
-    "milesstones": {
+    "milestones": {
```
webpro, line 31: ❤️

webpro (approved):
Thanks, @jeffesquivels!

I might toggle the default to `true` in v18.

webpro:
:rocket: _This pull request is included in v17.7.0. See [Release 17.7.0](https://github.com/release-it/release-it/releases/tag/17.7.0) for release notes._

PR #1116, opened 2024-06-17
feat: create tag via Gitlab release (#1115)

afrimberger:
Create tag via Gitlab release #1115

Issue #1115, opened 2024-06-14
Gitlab: Create release without requiring git push


afrimberger:
**Current behavior**

Creating a Gitlab release without previous `git push` fails with `Response code 422 (Unprocessable Entity)` even though the Gitlab [releases API](https://docs.gitlab.com/ee/api/releases/#create-a-release) can create the corresponding tag automatically if it doesn't exist, yet (by passing `ref`).

**Motivation**

Using only the Gitlab API is much easier because you can just use the default token injected into the build job (`CI_JOB_TOKEN`). For using `git push` you need to create an Access Token what is cumbersome.

**Example configuration**

```json
{
  "git": {
    "push": false,
    "tag": false,
    "commit": false,
    "requireCleanWorkingDir": true
  },
  "gitlab": {
    "release": true,
    "tokenRef": "CI_JOB_TOKEN",
    "tokenHeader": "Job-Token"
  }
}
```

**What needs to be changed**

Just pass `branchName` in the field `ref` in the `createRelease()` method. Two lines in `GitLab.js` need to be adapted:

[Line 184](https://github.com/release-it/release-it/blob/3056a184265d60487ea570f85cfcd9835c6f8bb6/lib/plugin/gitlab/GitLab.js#L184):
```javascript
const { tagName, branchName } = this.config.getContext();
```

and [Line 200](https://github.com/release-it/release-it/blob/3056a184265d60487ea570f85cfcd9835c6f8bb6/lib/plugin/gitlab/GitLab.js#L200):
```javascript
const options = {
    json: {
        name,
        ref: branchName,  // pass branch name
        tag_name: tagName,
        description
    }
};
```

webpro:
Thanks @afrimberger, feel free to open a pull request and I'm happy to look into it.

fjakop:
We would appreciate this also very much, please merge this.

webpro:
Can't merge, unfortunately the PR contains issues.

robross0606:
> Can't merge, unfortunately the PR contains issues.

@webpro, what issues?

webpro:
https://github.com/release-it/release-it/pull/1116#issuecomment-2182254217

Feel free to open a new PR though, seems like a fairly doable feat.

robross0606:
You would think so, but the test feedback is almost non-existent.  What is failing?  The MR changes are so minimal, I can't fathom what's failing.

webpro:
[...]

PR #1161, opened 2024-10-11
feat: create tag via Gitlab release (#1115)

robross0606:
Replacement for release-it#1116  as per https://github.com/release-it/release-it/pull/1116#issuecomment-2405765425 with:

- Functional unit tests. 🤞 
- Support for `tag_message` to [GitLab API](https://docs.gitlab.com/ee/api/releases/#create-a-release) via `git.tagAnnotation`.

### DIFF
--- a/test/gitlab.js
+++ b/test/gitlab.js
@@ -2,6 +2,8 @@ import fs from 'node:fs';
 import test from 'ava';
 import sinon from 'sinon';
 import nock from 'nock';
+import { isCI } from 'ci-info';
+import Git from '../lib/plugin/git/Git.js';
 import GitLab from '../lib/plugin/gitlab/GitLab.js';
 import { factory, runTasks } from './util/index.js';
 import {
@@ -63,6 +65,9 @@ test.serial('should upload assets and release', async t => {
   const gitlab = factory(GitLab, { options });
   sinon.stub(gitlab, 'getLatestVersion').resolves('2.0.0');
 
+  const git = factory(Git, { options });
+  const branchName = await git.getBranchName();
+
   interceptUser();
   interceptCollaborator();
   interceptMilestones({
@@ -89,7 +94,9 @@ test.serial('should upload assets and release', async t => {
   interceptPublish({
     body: {
       name: 'Release 2.0.1',
+      ref: isCI ? 'HEAD' : branchName,
       tag_name: '2.0.1',
+      tag_message: 'Release 2.0.1',
       description: 'Custom notes',
       assets: {
         links: [
--- a/test/tasks.interactive.js
+++ b/test/tasks.interactive.js
@@ -171,7 +171,9 @@ test.serial('should run "after:*:release" plugin hooks', async t => {
     project,
     body: {
       name: 'Release 1.1.0',
+      ref: 'master',
       tag_name: '1.1.0',
+      tag_message: 'Release 1.1.0',
       description: `* More file (${sha})`
     }
   });
--- a/test/tasks.js
+++ b/test/tasks.js
@@ -270,7 +270,9 @@ test.serial('should release all the things (pre-release, github, gitlab)', async
     project,
     body: {
       name: 'Release 1.1.0-alpha.0',
+      ref: 'master',
       tag_name: 'v1.1.0-alpha.0',
+      tag_message: `${owner} ${owner}/${project} ${project}`,
       description: `Notes for ${pkgName}: ${sha}`,
       assets: {
         links: [
--- a/lib/plugin/gitlab/GitLab.js
+++ b/lib/plugin/gitlab/GitLab.js
@@ -189,10 +189,11 @@ class GitLab extends Release {
 
   async createRelease() {
     const { releaseName } = this.options;
-    const { tagName } = this.config.getContext();
+    const { tagName, branchName, git: { tagAnnotation } = {} } = this.config.getContext();
     const { id, releaseNotes, repo, origin } = this.getContext();
     const { isDryRun } = this.config;
     const name = format(releaseName, this.config.getContext());
+    const tagMessage = format(tagAnnotation, this.config.getContext());
     const description = releaseNotes || '-';
     const releaseUrl = `${origin}/${repo.repository}/-/releases`;
     const releaseMilestones = this.getReleaseMilestones();
@@ -208,7 +209,9 @@ class GitLab extends Release {
     const options = {
       json: {
         name,
+        ref: branchName,
         tag_name: tagName,
+        tag_message: tagMessage,
         description
       }
     };
--- a/package-lock.json
+++ b/package-lock.json
@@ -6384,11 +6384,11 @@
       ]
     },
     "node_modules/micromatch": {
-      "version": "4.0.5",
-      "resolved": "https://registry.npmjs.org/micromatch/-/micromatch-4.0.5.tgz",
-      "integrity": "sha512-DMy+ERcEW2q8Z2Po+WNXuw3c5YaUSFjAO5GsJqfEl7UjvtIuFKO6ZrKvcItdy98dwFI2N1tg3zNIdKaQT+aNdA==",
+      "version": "4.0.8",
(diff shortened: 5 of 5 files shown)


package.json
```diff
@@ -140,5 +140,8 @@
     "plugins": [
       "preset-webpro"
     ]
+  },
+  "volta": {
+    "node": "18.20.4"
```
webpro, line 145: I'm OK with including this, but would it make sense to use the same range as `engines.node`?

test/tasks.js
```diff
@@ -237,7 +237,7 @@ test.serial('should release with correct tag name', async t => {
   exec.restore();
 });
 
-test.serial('should release all the things (pre-release, github, gitlab)', async t => {
+test.serial.only('should release all the things (pre-release, github, gitlab)', async t => {
```
webpro, line 240: Please remove `.only`

webpro (approved):
Thanks @robross0606. LGTM, just two small things.

package.json
robross0606: Oops, sorry!  I didn't mean to commit that.  `volta` doesn't support ranges.  That was meant purely for my local config.  I will remove.

robross0606: Removed.

test/tasks.js
robross0606: Removed

robross0606:
Well I'll be...  It is failing in some scenarios but not failing locally.  I don't have all the alternate OS to test this.

test/gitlab.js
```diff
@@ -89,7 +89,9 @@ test.serial('should upload assets and release', async t => {
   interceptPublish({
     body: {
       name: 'Release 2.0.1',
+      ref: 'feature/gitlab-release-with-ref',
```
robross0606, line 92: This is causing a unit test problem since `branchName` resolves differently when tests are run locally than when being run on the server.  Locally this resolves to `'feature/gitlab-release-with-ref'` but it is resolving to `'HEAD'` when run on GitHub.

robross0606: @webpro, I think I've reached the limit of how I can help here without some guidance.  I'm having a heckuva time trying to figure out how to solve this.  There's very little inline documentation, and I'm unfamiliar with several of the core libraries being used.

robross0606: Unless I'm mistaken, one of the key differences in the `gitlab.js` tests is that it does not start by initializing an empty git repository from which to run tests.  Both `tasks.interactive.js` and `tasks.js` do this with `beforeEach()` but `gitlab.js` does not.  Because of this, the tests in this suite are not deterministic because `branchName` would evaluate differently for every branch in which the tests are run.  This feels like a more fundamental problem with the unit test setup than I am equipped to resolve and somewhat outside the scope of this specific ticket.

robross0606: One thing we _could_ do here to side-step the issue entirely is to change the `nock` body matching for the `ref` property to some type of regex for that specific test:
image
I don't know how you'd feel about that.  It feels a little like cheating. 🙂

test/tasks.interactive.js
```diff
@@ -171,7 +171,9 @@ test.serial('should run "after:*:release" plugin hooks', async t => {
     project,
     body: {
       name: 'Release 1.1.0',
+      ref: 'master',
```
webpro, line 174:
```suggestion
      ref: isCI ? 'HEAD' : 'main',
```

test/tasks.js
```diff
@@ -270,7 +270,9 @@ test.serial('should release all the things (pre-release, github, gitlab)', async
     project,
     body: {
       name: 'Release 1.1.0-alpha.0',
+      ref: 'master',
```
webpro, line 273:
```suggestion
      ref: isCI ? 'HEAD' : 'main',
```

test/gitlab.js
webpro: Wouldn't be the end of the world if required, but let's try my suggestion first :)

webpro (commented):
Thanks for all the efforts you've put into this! Let's see if we can wrap this up.

test/gitlab.js
robross0606: Which suggestion?  This definitely won't continue working as-is because I currently have the name of my current branch in there.  If I were to leave it as-is, it would eventually fail on someone else's branch with a different name.  One other thing we could do is detect the current branch's name and use that in lieu of my hard-coded one in your `isCI` check.

test/tasks.interactive.js
robross0606: Where does `isCI` come from?  I don't see that as being a valid variable.  I do see a function of that name buried in various `config` objects in the test, not not directly as a free-standing variable.

webpro: Oh dang, auto-import when I type in IDE:

```
import { isCI } from 'ci-info';
```

robross0606: Because this appears to create a new "bare" repo and use that as the basis for further testing, I think this one will always be "master" no matter what.  It is only the `gitlab.js` tests that having this problem because it doesn't test within a temporary bare repo.

test/tasks.js
robross0606: Because this appears to create a new "bare" repo and use that as the basis for further testing, I think this one will always be "master" no matter what.  It is only the `gitlab.js` tests that having this problem because it doesn't test within a temporary bare repo.

test/tasks.interactive.js
webpro: Normally I'd push to your branch but somehow that isn't working now.

test/tasks.js
webpro: The default on both GitHub and my machine is "main".

webpro: And the test in the CI test runner env says `HEAD`.

robross0606: Interesting.  This is _very_ envirionment-specific.

webpro: We're using any string on any machine=, but in CI a fresh checkout is always `HEAD` if I'm not mistaken. If my assumption is incorrect we can just always use the regex match.

robross0606: This does not appear to be correct.  All the places I changed to `'HEAD'` are now failing because they are being issued as `'master'`.  `'HEAD'` is a not really a branch name.  It is a shortcut to the latest commit.  It might be evaluating that way in CI in certain places if you're checking out "untracked" which means the working code is not related to any specific upstream branch.  However, for the two tests on `tasks` and `tasks.interactive` those always appear to be `'master'` (at least locally an on CI) because they're operating from a completely independent bare repository which is created in a temporary directory.  In other words, those two sets of tests don't operate directly on the `release-it`'s repository while `gitlab` tests do.

webpro: >  those two sets of tests don't operate directly on the `release-it`'s repository while `gitlab` tests do

I totally missed that the GitLab tests do this. Sorry, been a while working on those. That's pretty bad actually. Still, I would not expect `actions/checkout` to create a `master` branch (https://github.com/release-it/release-it/actions/runs/11333030713/job/31516426458#step:2:83) so I guess we're operating on it from the tests. Thanks for bearing with me :)

robross0606: `actions/checkout` is not creating a master branch.  If you look, those two tests use the `gitAdd()` utility function to create a whole separate "bare" repository from which to work.  So it is basically creating an entirely separate bogus repo from which to run tests.  This seems like a much better and controlled approach than what the `gitlab.js` tests are doing which relies on whatever the environment happens to be doing with the `release-it` repo itself.

webpro: Yes, I think we're saying the same.

robross0606: Unless I'm mistaken (GitHub workflow is still a little new to me), I think all tests have now passed.

webpro:
:rocket: _This pull request is included in v17.9.0. See [Release 17.9.0](https://github.com/release-it/release-it/releases/tag/17.9.0) for release notes._

PR #1166, opened 2024-10-16
Add proper Content-Type header to fetch requests

jeffesquivels:
The `got` npm package [sets the proper](https://github.com/sindresorhus/got/blob/main/documentation/quick-start.md?plain=1#L60) `Content-Type` header when the body of a request is set to `json`. This behavior [was not ported](https://github.com/release-it/release-it/commit/696f0821fa871a794dba9eabdcc960554f073eab#diff-44f297bf9ddf782cf7057c9b2755dedcb77fd7e168cbd354690eb933a9b5022bR166-R169) when `got` and `node-fetch` were removed in favor of NodeJS' native `fetch` API.

This PR fixes this.


webpro:
Thanks @jeffesquivels! Happy to merge. Just wonder if `text/plain` is the best fallback. I will pick up #1167 so we'll see :)

PR #1169, opened 2024-10-18
Feat: Add `publishPackageManager` config option to allow using different package manager for publishing (e.g. Bun)

chrispader:


webpro:
Thanks for the PR, Christoph! We've tried this before with Yarn and pnpm but remember not adding support for various reasons (could look them up in this repo issue history).

So I think we could do something like this, but only for Bun and only if there's no issues around still using `npm version` + `bun publish` (e.g. `workspace:` protocol comes to mind). I'll look into it a bit more when I find the time.

qodesmith:
Perhaps the `Config` type can be updated and `publishPackageManager` can be `'npm' | 'bun'` to indicate these are the only two supported package managers at this time.

webpro:
What this PR needs is a bit of confidence that using `npm` for `npm version` does not cause issues when Bun is selected, or update that command as well to use `bun` instead. A test or two to verify things wouldn't hurt either.

kathleenmotley49-maker (approved):

webpro:
Example usage:

- https://github.com/webpro-nl/knip/blob/main/.release-it.json
- https://github.com/webpro-nl/knip/blob/main/release.sh

Using release-it there to release all packages in monorepo in one go locally, seems to work well, including shared root `.release-it.json` config.

chrispader:
Thanks @webpro, sorry i didn't get to do further testing on this. Glad this works! 🙌🏼

PR #1181, opened 2024-12-08
feat: add capability to add extends configuration to release-it

### LINKED ISSUE #1180: Extendable configuration
Following the idea of https://github.com/release-it/release-it/issues/934, let's add as the first extendable configuration to the following:

1. Extend the configuration using the `extends` and merge the local configuration with the remote configuration
2. The local will always take precedence over the extends configuration
3. The extends must be accessible via the internet

Note: I think as part of the first iteration, public access is enough to make it work, and then we can add authorization to make it work behind a private access repo.

juancarlosjr97:
This pull request introduces the ability to extend configuration from a remote source, specifically GitHub repositories, and includes several related changes to the codebase, documentation, and tests.

### New Feature: Extend Configuration

* [`docs/configuration.md`](diffhunk://#diff-17ed18489a956f326ec0fe4040850c5bc9261d4631fb42da4c52891d74a59180R90-R108): Added documentation for the new `extends` option that allows extending configurations from remote GitHub repositories.
* [`schema/release-it.json`](diffhunk://#diff-351d879d6af18de4b75acfc02e1d443c8ffb3438f9d1176537f8eafeb7e46f4dR12-R15): Added the `extends` property to the JSON schema to specify a configuration to extend.

### Code Changes

* [`lib/config.js`](diffhunk://#diff-d5109920b34c2b2e141bd41232d8876432a17479c388e142d8a036bb7b36ea62L6-R6): Introduced `fetchConfigurationFromGitHub` and `getRemoteConfiguration` functions to fetch and merge remote configurations. Updated the `Config` class to include `mergeRemoteOptions` method. [[1]](diffhunk://#diff-d5109920b34c2b2e141bd41232d8876432a17479c388e142d8a036bb7b36ea62L6-R6) [[2]](diffhunk://#diff-d5109920b34c2b2e141bd41232d8876432a17479c388e142d8a036bb7b36ea62R41-R67) [[3]](diffhunk://#diff-d5109920b34c2b2e141bd41232d8876432a17479c388e142d8a036bb7b36ea62R113-R116) [[4]](diffhunk://#diff-d5109920b34c2b2e141bd41232d8876432a17479c388e142d8a036bb7b36ea62R172-R173)
* [`lib/index.js`](diffhunk://#diff-92bbac9a308cd5fcf9db165841f2d90ce981baddcb2b1e26cfff170929af3bd1L4-R4): Updated the `runTasks` function to fetch and merge remote configurations if the `extends` property is present. [[1]](diffhunk://#diff-92bbac9a308cd5fcf9db165841f2d90ce981baddcb2b1e26cfff170929af3bd1L4-R4) [[2]](diffhunk://#diff-92bbac9a308cd5fcf9db165841f2d90ce981baddcb2b1e26cfff170929af3bd1R17-R25)

### Tests

* [`test/config.js`](diffhunk://#diff-324e5ae6ed1d64846d943bc35a8a7394f2099af356ca5d9e30413648a3ae124fL3-R17): Added tests to verify fetching extended configurations from GitHub with various scenarios (default file, specific tag, custom file). [[1]](diffhunk://#diff-324e5ae6ed1d64846d943bc35a8a7394f2099af356ca5d9e30413648a3ae124fL3-R17) [[2]](diffhunk://#diff-324e5ae6ed1d64846d943bc35a8a7394f2099af356ca5d9e30413648a3ae124fR167-R274)
* [`test/tasks.interactive.js`](diffhunk://#diff-d71ed2d617b85d6178e3de2af9db613af912bb4ea00248178ed8d868e21d710aR93-R153): Added tests to ensure tasks run correctly with and without extended configurations.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -91,16 +91,16 @@ Release-it has [plenty of options][2]. See the following tables for plugin confi
 
 You can extend a configuration from a remote source using the `extends` option. The following formats are supported:
 
-- `github>owner/repo`
-- `github>owner/repo#tag`
-- `github>owner/repo:file#tag`
+- `github:owner/repo`
+- `github:owner/repo#tag`
+- `github:owner/repo:file#tag`
 
 For example, to extend a configuration from a GitHub repository:
 
 ```json
 {
-  "$schema": "https://unpkg.com/release-it@17/schema/release-it.json",
-  "extends": "github>release-it/release-it-configuration"
+  "$schema": "https://unpkg.com/release-it@18/schema/release-it.json",
+  "extends": "github:release-it/release-it-configuration"
 }
 ```
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -87,6 +87,25 @@ Release-it has [plenty of options][2]. See the following tables for plugin confi
 - [GitHub][5]
 - [GitLab][6]
 
+### Extend Configuration
+
+You can extend a configuration from a remote source using the `extends` option. The following formats are supported:
+
+- `github>owner/repo`
+- `github>owner/repo#tag`
+- `github>owner/repo:file#tag`
+
+For example, to extend a configuration from a GitHub repository:
+
+```json
+{
+  "$schema": "https://unpkg.com/release-it@17/schema/release-it.json",
+  "extends": "github>release-it/release-it-configuration"
+}
+```
+
+Only public GitHub repositories are supported at this time.
+
 ## Setting options via CLI
 
 Any option can also be set on the command-line, and will have highest priority. Example:
--- a/lib/config.js
+++ b/lib/config.js
@@ -41,17 +41,17 @@ const getLocalConfig = ({ file, dir = process.cwd() }) => {
   return result && isObjectStrict(result.config) ? result.config : localConfig;
 };
 
-const fetchConfigurationFromGitHub = async pattern => {
+const fetchConfigurationFromGitHub = async configuration => {
   const regex = /^github:([^/]+)\/([^#:]+)(?::([^#]+))?(?:#(.+))?$/;
-  const match = pattern.match(regex);
+  const match = configuration.match(regex);
 
   if (!match) {
-    throw new Error('Invalid Extended Configuration from GitHub');
+    throw new Error(`Invalid Extended Configuration from GitHub: ${configuration}`);
   }
 
   const [, owner, repo, file = '.release-it.json', tag] = match;
-  const branchOrTag = tag ? `refs/tags/${tag}` : 'HEAD';
-  const url = `https://raw.githubusercontent.com/${owner}/${repo}/${branchOrTag}/${file}`;
+  const ref = tag ? `refs/tags/${tag}` : 'HEAD';
+  const url = new URL(`https://raw.githubusercontent.com/${owner}/${repo}/${ref}/${file}`);
 
   const response = await fetch(url);
 
--- a/lib/config.js
+++ b/lib/config.js
@@ -6,7 +6,7 @@ import defaultsDeep from '@nodeutils/defaults-deep';
 import { isObjectStrict } from '@phun-ky/typeof';
 import merge from 'lodash.merge';
 import get from 'lodash.get';
-import { readJSON, getSystemInfo } from './util.js';
+import { e, getSystemInfo, readJSON } from './util.js';
 
 const debug = util.debug('release-it:config');
 const defaultConfig = readJSON(new URL('../config/release-it.json', import.meta.url));
@@ -42,11 +42,13 @@ const getLocalConfig = ({ file, dir = process.cwd() }) => {
 };
 
 const fetchConfigurationFromGitHub = async configuration => {
+  const docs = 'https://github.com/release-it/release-it/blob/main/docs/configuration.md';
+
   const regex = /^github:([^/]+)\/([^#:]+)(?::([^#]+))?(?:#(.+))?$/;
   const match = configuration.match(regex);
 
   if (!match) {
-    throw new Error(`Invalid Extended Configuration from GitHub: ${configuration}`);
+    throw e(`Invalid Extended Configuration from GitHub: ${configuration}`, docs);
   }
 
   const [, owner, repo, file = '.release-it.json', tag] = match;
@@ -56,7 +58,7 @@ const fetchConfigurationFromGitHub = async configuration => {
   const response = await fetch(url);
 
   if (!response.ok) {
-    throw new Error(`Failed to fetch ${url}: ${response.statusText}`);
+    throw e(`Failed to fetch ${url}: ${response.statusText}`, docs);
   }
 
   return response.json();
--- a/lib/config.js
+++ b/lib/config.js
@@ -38,8 +38,32 @@ const getLocalConfig = ({ file, dir = process.cwd() }) => {
     throw new Error(`Invalid configuration file at ${result.filepath}`);
   }
   debug({ cosmiconfig: result });
-
   return result && isObjectStrict(result.config) ? result.config : localConfig;
+};
+
+const fetchConfigurationFromGitHub = async pattern => {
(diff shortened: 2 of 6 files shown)


juancarlosjr97:
I know you probably are busy @webpro, but if you have some time, any feedback would be great ;)

webpro:
Thanks for the PR, Juan.

While I appreciate the convenience this PR will bring, it feels odd to fetch a configuration from GitHub like that.  

Wouldn't it be less involved and less risky if we would accept only dynamic `.js` or `.ts` configurations, something like this:

```ts 
// .release-it.ts
import extConfig from 'release-it-configuration';

const localConfig = {};

export default {
  ...extConfig,
  ...localConfig
}
```

This basically does not need a PR at all, but what would be the downside of it? It's also more flexible to adjust to whatever is needed. And we don't need to be bothered with authorization (something you mentioned in the other issue).

It's also possible to install packages from github directly if that's what's needed (e.g. using the `github:user/repo` syntax).

Overall I think I'm not sure whether we should add this complexity to release-it, while users can share configurations already through npm packages directly.

juancarlosjr97:
Thanks @webpro  :) 

> Wouldn't it be less involved and less risky if we would accept only dynamic .js or .ts configurations

Yes, I agree it would be less risky given we are merging the configurations.

> This basically does not need a PR at all, but what would be the downside of it? It's also more flexible to adjust to whatever is needed. And we don't need to be bothered with authorization (something you mentioned in the other issue).

The main reason I think is good to have it external and be fetched by release-it is to avoid the complexity of adding packages, given that not all projects that use release-it might not have npm or any other package manager to install the dependencies.

> It's also possible to install packages from github directly if that's what's needed (e.g. using the github:user/repo syntax).

I did not know this! I think that helps to make it easier, but still needs the package.json and requires project to have also a javascript package manager to install the dependencies.

----

Given that release-it, even though it is JavaScript-based, can be used in any context, I think it could handle the fetch and merge in the code.

However, my objective is for people to use, shared configurations, to keep it as part of release-it so others can use it too without the need to build something around to fetch and get the configuration.

juancarlosjr97:
I think the most important consideration is authentication for private endpoints, which honestly is the aspect that I do not like :(

lib/index.js
```diff
@@ -14,6 +14,15 @@ const runTasks = async (opts, di) => {
     Object.assign(container, di);
     container.config = container.config || new Config(opts);
 
+    if ('extends' in container.config.options) {
```
webpro, line 17: Perhaps `typeof 'string'` makes sense, so you can set it to `false` or `undefined` as well to disable without issues (there are types etc. but they're not always used/adhered to).

webpro (commented):
Appreciate you sticking with it, and having proper tests!

> The main reason I think is good to have it external and be fetched by release-it is to avoid the complexity of adding packages, given that not all projects that use release-it might not have npm or any other package manager to install the dependencies.

Very good point, my JS brain totally forgot about that for a minute.

> I think the most important consideration is authentication for private endpoints

Perhaps we can get away with it by stating that _shareable_ configs must be _public_, at least initially. Doesn't sounds very unreasonable to me :)

lib/index.js
juancarlosjr97: I like that! I will update it alone the tests :)

juancarlosjr97: Updated with tests :)

juancarlosjr97: Updated with tests :)

docs/configuration.md
```diff
@@ -87,6 +87,25 @@ Release-it has [plenty of options][2]. See the following tables for plugin confi
 - [GitHub](./github-releases.md#configuration-options)
 - [GitLab](./gitlab-releases.md#configuration-options)
 
+### Extend Configuration
+
+You can extend a configuration from a remote source using the `extends` option. The following formats are supported:
+
+- `github>owner/repo`
+- `github>owner/repo#tag`
+- `github>owner/repo:file#tag`
```
webpro, line 96: How do you feel about using `:` as more of a "protocol delimiter"?

Can't remember having seen `>` myself before, and `:` is more common?

By the way we could also use something like 

- `npm:juancarlosjr97/release-it-configuration/release-it.json` and/or even
- `https://unpkg.com/juancarlosjr97/release-it-configuration/schema/release-it.json` 

for anything, also anything published on npm is accessible e.g. https://unpkg.com/release-it@17.10.0/config/release-it.json.

lib/config.js
```diff
@@ -38,6 +38,31 @@ const getLocalConfig = ({ file, dir = process.cwd() }) => {
   return result && _.isPlainObject(result.config) ? result.config : localConfig;
 };
 
+const fetchConfigurationFromGitHub = async pattern => {
+  const regex = /^github>([^/]+)\/([^#:]+)(?::([^#]+))?(?:#(.+))?$/;
+  const match = pattern.match(regex);
+
+  if (!match) {
+    throw new Error('Invalid Extended Configuration from GitHub');
```
webpro, line 46: This could perhaps be improved, maybe add `github` as `protocol`/(named) match (like `owner` etc. below). Also please look for other instances of `throw e()`.

docs/configuration.md
juancarlosjr97: I am not sure if I have seen a format like this anywhere else besides [Renovate Preset File Naming](https://docs.renovatebot.com/config-presets/#preset-file-naming). It follows the same format I proposed, which is why I went with it. That said, I am happy to change it, as it is ultimately just an abstract representation.

Having written that, I think `:` has a stronger logical association with URLs compared to `>` now that you mention it, so I will change it :) 

> By the way, we could also use something like...

I really like the idea of using `unpkg`. That way, we are not constrained to GitHub (only to its other owned package registry 😄) 

Would you like me to work on adding npm (unpkg) in the current PR, or should I raise another issue and tackle it as a separate PR? I am happy to do it either way!

lib/config.js
juancarlosjr97: Honestly, I have never been a big fan of regex, I think I prefer JSONana to extract information, and build logical around it.

I will update the error following other patterns!

Thanks!

webpro: Maybe we could parse it as a `new URL()`

docs/configuration.md
webpro: Happy to leave that up to you. Maybe good to keep this one small, but if another PR's overhead you can add it here too.

juancarlosjr97: Update to use the delimiter between github and the path to `:`

juancarlosjr97:
@webpro sorry for the super delay, literally over 2 months, but I had some weekend time to wrap this up :)

juancarlosjr97:
Whenever you have a chance @webpro! Thank you!!

juancarlosjr97:
I need to update the tests, it looks like rebased caused a failure :(

juancarlosjr97:
All passing now @webpro  :)

webpro:
Rebased on top of recent merges to `main` (#1201 and #1202)

lib/config.js
juancarlosjr97: Sure, sounds good! I will aim to do the changes before the end of the week!

juancarlosjr97: @webpro I did not do these changes. I did attempt but I had to think about the tag reference, I was going to get intro substring extraction. 

I can happily raise another issue for an improvement for the configuration, as this is already released.

juancarlosjr97:
Thank you @webpro. Let me know about the comment and I can raise an issue for the improvement of the details configuration extraction.

webpro:
Oops, didn't realize. It got merged - also as a _driver_ to get things going - but not actually released (other than [through pkg.pr.new](https://github.com/release-it/release-it/actions/runs/13946108795/job/39033603324)).

Let's get that config url handling fixed before v19 goes out the door (we're still in a somewhat lengthy prerelease phase).

webpro:
:rocket: _This pull request is included in v19.0.0-next.2. See [Release 19.0.0-next.2](https://github.com/release-it/release-it/releases/tag/19.0.0-next.2) for release notes._

PR #1183, opened 2024-12-09
fix(gitlab): don't set Content-Type header when uploading release assets

### LINKED ISSUE #1172: gitlab release fails to upload
So I'm trying to create a gitlab release, but it throws the following error:
```
ERROR The provided content-type 'text/plain' is not supported.
```

This is my configuration:
```json
{
    "$schema": "https://unpkg.com/release-it/schema/release-it.json",
    "git": {
        "commitMessage": "chore: Release v${version}"
    },
    "gitlab": {
        "assets": ["dist/*"],
        "release": true,
        "useIdsForUrls": true
    },
    "hooks": {
        "before:git:release": "yarn build"
    },
    "plugins": {
        "@release-it/conventional-changelog": {
            "infile": "CHANGELOG.md",
            "preset": {
                "name": "conventionalcommits"
            }
        }
    }
}
``` 

And this is what i executed:
```sh
NODE_DEBUG=release-it:* GITLAB_TOKEN=<token> yarn release -VVV --no-plugins.@release-it/conventional-changelog.infile --no-increment --no-git --gitlab.release --gitlab.assets="dist/*"
```
[output.txt](https://github.com/user-attachments/files/17509809/output.txt)

---

So it does try to upload the 4 files, but gitlab doesn't seem to like the `content-type`.

---

Some relevant code:
https://github.com/release-it/release-it/blob/28213bf51d80e9cb9157619a72e81595879d4610/lib/plugin/gitlab/GitLab.js#L254-L276
https://github.com/release-it/release-it/blob/28213bf51d80e9cb9157619a72e81595879d4610/lib/plugin/gitlab/GitLab.js#L168

rjw57:
As noted in [a comment on #1172][1], switching to the use of the native fetch API broke uploading release assets to GitLab by hard-coding the Content-Type header to "text/plain". The [recommended approach][2] is to avoid setting the Content-Type header at all and let the fetch API implementation set it so as to ensure that the correct boundary expression is included.

~~Add a flag to the `request` function to allow explicitly disabling setting the Content-Type header.~~ Update the `request` function to avoid setting the Content-Type header if the body is an instance of `FormData`. The default behaviour is to set the header as before but we disable setting the header only in the `uploadAsset` function.

The bug is blocking a [feature in our product](https://gitlab.developers.cam.ac.uk/uis/devops/iam/activate-account/api/-/issues/30) and I've checked that the change in this PR fixes our observed issue with GitLab release asset uploads. Using the branch from this PR, I am able to create a release with the expected assets in a test project.

[1]: https://github.com/release-it/release-it/issues/1172#issuecomment-2468592425
[2]: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest_API/Using_FormData_Objects#sending_files_using_a_formdata_object

Closes #1172

### DIFF
--- a/lib/plugin/gitlab/GitLab.js
+++ b/lib/plugin/gitlab/GitLab.js
@@ -165,9 +165,13 @@ class GitLab extends Release {
     const url = `${baseUrl}/${endpoint}${options.searchParams ? `?${new URLSearchParams(options.searchParams)}` : ''}`;
     const headers = {
       'user-agent': 'webpro/release-it',
-      'Content-Type': typeof options.json !== 'undefined' ? 'application/json' : 'text/plain',
       [tokenHeader]: this.token
     };
+    // When using fetch() with FormData bodies, we should not set the Content-Type header.
+    // See: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest_API/Using_FormData_Objects#sending_files_using_a_formdata_object
+    if (!(options.body instanceof FormData)) {
+      headers['Content-Type'] = typeof options.json !== 'undefined' ? 'application/json' : 'text/plain';
+    }
     const requestOptions = {
       method,
       headers,


lib/plugin/gitlab/GitLab.js
```diff
@@ -260,7 +263,12 @@ class GitLab extends Release {
     const body = new FormData();
     const rawData = await fs.promises.readFile(filePath);
     body.set('file', new Blob([rawData]), name);
-    const options = { body };
+    const options = {
+      body,
+      // When using fetch() with FormData bodies, we should not set the Content-Type header.
+      // See: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest_API/Using_FormData_Objects#sending_files_using_a_formdata_object
+      setContentType: false
```
webpro, line 270: Maybe we can check for something like `body instanceof FormData` in the `this.request` method without having to pass this option?

webpro (commented):
Thanks @rjw57, much appreciated! Happy to merge, but I do have a question.

lib/plugin/gitlab/GitLab.js
rjw57: We certainly could. This would tend to increase the "magic" in the request function and my taste is to be explicit rather than implicit but it's also not my project. Would you prefer to take that approach? If so, let me know and I'll re-work.

rjw57: No problem - I'll re-work.

rjw57: That's now been done. The diff is indeed now a lot neater.

rjw57: I take your point and am always happy to adjust style to match the project.

webpro:
:rocket: _This pull request is included in v18.0.0-next.4. See [Release 18.0.0-next.4](https://github.com/release-it/release-it/releases/tag/18.0.0-next.4) for release notes._

webpro:
:rocket: _This pull request is included in v17.11.0. See [Release 17.11.0](https://github.com/release-it/release-it/releases/tag/17.11.0) for release notes._

Issue #1184, opened 2024-12-10
Regression in self-hosted gitlab releases after switching to native fetch


vonBrax:
Hi again!

A while ago I submitted a PR #1129  to fix #1125 to allow insecure connections to self hosted gitlab instances.
The problem is back again after removing `node-fetch` and `got` in favor of native `fetch` on `v17.7.0` .
Additionally, the `certificateAuthorityFile` option also no longer works.

To make these settings work with the native `fetch`, we need to pass down an `Agent` instance from [undici](https://github.com/nodejs/undici) to the `fetch` ([see comment](https://github.com/nodejs/node/issues/48977#issuecomment-1658324420)).

`Node` uses `undici` internally for the fetch requests and currently does not publicly expose it, but there is an [open issue](https://github.com/nodejs/node/issues/43187) to expose some of its APIs.

I could submit a PR to update the `fetch` configuration to make these settings work again, but that would mean that we would need to add `undici` as a new dependency, at least temporarily until `node` exposes the `Agent` class.

If that would be an acceptable solution, let me know and I'll create the PR.

I'm also open to suggestions if there are any other better solutions.

webpro:
Sorry for the hassle, it was a bit unfortunate without having tests for the case. A PR for this would be fine with me.

vonBrax:
Yeah, it would be nice if we could actually test these settings, but I'm not sure if that's something that could easily be done. I'll see if I can come up with something, maybe spinning up a local server or something.
I'll prepare the PR.
Thank you for your prompt answer!

webpro:
The issue seems to have been addressed, closing this one.

PR #1185, opened 2024-12-11
fix: pass gitlab options to fetch agent

vonBrax:
Fix: #1184 

This PR fix the issues with the gitlab `secure` and `certificateAuthorityFile` configuration options.

Notes:
 - Added `undici` as dependency
 - `undicy` >= v7.0.0 has a more restrict node version requirement: ">=20.18.1"
 - Current `release-it` node requirement: "^18.18.0 || ^20.9.0 || ^22.0.0"

**Edit:**
 - `undici` v6.21.0 supports node 18 --> downgraded


vonBrax:
@webpro

Could you please review this PR?

Also, keep in mind that this PR could introduce breaking changes for people using node < 20.18.1.
I couldn't really test whether it would still work for these older node versions.

webpro:
You could also consider a PR against the `v18` branch, to support Node.js v20+. I've been a bit slow with releasing that, could be done this month. When this PR gets merged, you can use `release-it@next` until v18 `latest` is out.

lib/plugin/gitlab/GitLab.js
```diff
@@ -17,15 +18,14 @@ class GitLab extends Release {
...
-    const httpsOptions = {
-      certificateAuthority: certificateAuthorityFile ? fs.readFileSync(certificateAuthorityFile) : undefined,
-      rejectUnauthorized: typeof secure === 'boolean' ? secure : undefined
-    };
-
-    // Remove keys with undefined values
-    const https = _.pickBy(httpsOptions, value => value !== undefined);
-
-    this.certificateAuthorityOption = _.isEmpty(https) ? {} : { https };
+    this.certificateAuthorityOption = {
```
webpro, line 21: Is it possible to have `this.certificateAuthorityOption` be an empty object as to minimize risks for existing users than don't need the custom `Agent`?

webpro (commented):
Good stuff, happy to merge. One question tho.

webpro:
Oh please run `npm run format`

lib/plugin/gitlab/GitLab.js
vonBrax: yep, no problems, I'll update it

vonBrax:
It seems like tests are failing under windows node v22, I'll take another look

vonBrax:
@webpro

it was a bit tricky to make the test server work under win node v22, because localhost would always resolve to an ipv6 address, while the server was listening on an ipv4 address. In the end I just stopped passing the host option to `server.listen`, which basically makes the server listen both on ipv4 and ipv6 unspecified ip addresses ([node documentation](https://nodejs.org/api/net.html#net_server_listen_port_host_backlog_callback)). It seems to be working now.

Anyway, should be good to go from my side now.

webpro:
Thanks Alexandre! Sorry for the trouble.

How can I test the cert script? On my machine, ` ./gen-cert.sh localhost` or `localhost:3000` doesn't seem to do the trick locally. What am I missing?

```
node -v
v22.12.0
```


Screenshot 2024-12-17 at 8 50 18 PM

vonBrax:
Hey man, no worries, no trouble at all. Sorry, I haven't really tested it on mac. I'm on ubuntu (WSL) also with node v22.12.0 and tests work locally for me, also after regenerating the certs.

 Can you try logging the actual error there ([this one here](https://github.com/release-it/release-it/blob/829ac382ab99a0b1278af71ca6ccb25da89c5779/lib/plugin/gitlab/GitLab.js#L71))?

I imagine it will be either a `ECONNREFUSED` or a `UNABLE_TO_VERIFY_LEAF_SIGNATURE`. If is the first one, then I might have a possible fix, but if is the latter, then I'll need to look into it.

I'll try to test it on a mac today to see if I can figure it out and I'll report it back.

webpro:
both tests fail with this error:

```
  [cause]: [Error: 800F61F901000000:error:0A00010B:SSL routines:ssl3_get_record:wrong version number:../deps/openssl/openssl/ssl/record/ssl3_record.c:355:
  ] {
    library: 'SSL routines',
    reason: 'wrong version number',
    code: 'ERR_SSL_WRONG_VERSION_NUMBER'
  }
```

is this about the openssl version on my machine perhaps? i have:

```
❯ openssl --version
OpenSSL 3.4.0 22 Oct 2024 (Library: OpenSSL 3.4.0 22 Oct 2024)
```

vonBrax:
Interesting... on the machine I'm using now I'm running a much older version than yours and it works:

`OpenSSL 3.0.2 15 Mar 2022 (Library: OpenSSL 3.0.2 15 Mar 2022)`

I just tested also on a Mac with `openSSL 3.3.6` and it also worked (current certs + regenerated). I'll see if I can update it to 3.4.0 to also test with that version.

I'm just gonna describe how I'm doing it, just to make sure we are on the same page:

```sh
cd test/util/https-server
./gen-cert.sh localhost
cd ../../..
npx ava test/gitlab.js
```

If you are doing these same steps and getting that error, then it's probably related to the openssl version and I don't know yet how to fix it, google searches didn't help much.

webpro:
yeah those were exactly the steps i took. the good news is that this works on my machine and tests pass:

- `brew install openssl@3.0` (can't rm or force-install a specific version otherwise)
- replace instances of `openssl` with `/opt/homebrew/opt/openssl@3.0/bin/openssl`
- looks like this is version 3.0.15

i'd consider this good enough, we can reproduce

vonBrax:
Ok, I just finished setting up a debian 11 docker container to test it with openssl 3.4.0 and my tests were still running fine there, so at this point I'm not sure what the issue was on your end, but it doesn't seem to be a problem exclusively from openssl.

Anyways, I think the fix present in this PR is fine, the problem is only with the tests. That means that we might run into troubles with the tests in the future...

Just as a note, I ran into some issues during the tests, but I later saw that the problem was that I was missing the rsync package - and I didn't immediately see that on the logs after running the gen-certs script. After installing the missing package and regenerating the certs, tests were successfully again.


image

webpro:
:rocket: _This pull request is included in v18.0.0-next.4. See [Release 18.0.0-next.4](https://github.com/release-it/release-it/releases/tag/18.0.0-next.4) for release notes._

Issue #1200, opened 2025-02-13
Please drop `shelljs` as a dep


phun-ky:
`shelljs` will not update to latest deps (vulnerability issues), and you might not need `shelljs` to traverse directories or to fire off child processes, node got that covered.

webpro:
Pull requests to phase out `shelljs` are most welcome! Any chance you're up for the task Alexander?

phun-ky:
I've already started, but it might need a separate sanity check from maintainers :P

PR #1201, opened 2025-02-14
fix: Remove `shelljs` as dependency, it causes vulnerability issues and is not…

phun-ky:
… needed. Closes: #1200

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -130,6 +130,8 @@ const tryStatFile = filePath => {
   }
 };
 
+const fixArgs = args => (args ? (typeof args === 'string' ? args.split(' ') : args) : []);
+
 export {
   getSystemInfo,
   format,
@@ -140,6 +142,7 @@ export {
   hasAccess,
   parseVersion,
   readJSON,
+  fixArgs,
   e,
   touch
 };
--- a/lib/plugin/git/Git.js
+++ b/lib/plugin/git/Git.js
@@ -2,14 +2,13 @@ import { EOL } from 'node:os';
 import _ from 'lodash';
 import { execa } from 'execa';
 import matcher from 'wildcard-match';
-import { format, e } from '../../util.js';
+import { format, e, fixArgs } from '../../util.js';
 import GitBase from '../GitBase.js';
 import prompts from './prompts.js';
 
 const noop = Promise.resolve();
 const invalidPushRepoRe = /^\S+@/;
 const options = { write: false };
-const fixArgs = args => (args ? (typeof args === 'string' ? args.split(' ') : args) : []);
 
 const docs = 'https://git.io/release-it-git';
 
--- a/lib/plugin/npm/npm.js
+++ b/lib/plugin/npm/npm.js
@@ -2,7 +2,7 @@ import path from 'node:path';
 import semver from 'semver';
 import urlJoin from 'url-join';
 import Plugin from '../Plugin.js';
-import { hasAccess, rejectAfter, parseVersion, readJSON, e } from '../../util.js';
+import { hasAccess, rejectAfter, parseVersion, readJSON, e, fixArgs } from '../../util.js';
 import prompts from './prompts.js';
 
 const docs = 'https://git.io/release-it-npm';
@@ -14,8 +14,6 @@ const DEFAULT_TAG = 'latest';
 const DEFAULT_TAG_PRERELEASE = 'next';
 const NPM_BASE_URL = 'https://www.npmjs.com';
 const NPM_PUBLIC_PATH = '/package';
-
-const fixArgs = args => (args ? (typeof args === 'string' ? args.split(' ') : args) : []);
 
 class npm extends Plugin {
   static isEnabled(options) {


phun-ky:
cc @webpro

lib/shell.js
```diff
@@ -1,12 +1,10 @@
 import util from 'node:util';
-import sh from 'shelljs';
+import sh from 'node:child_process';
 import { execa } from 'execa';
 import { format } from './util.js';
 
 const debug = util.debug('release-it:shell');
 
-sh.config.silent = !debug.enabled;
```
webpro, line 8: This functionality is pretty useful for a few reasons:

- Clean output in release-it output for users
- Clean output in release-it output during test runs

Output is disabled by default, and can be enabled using `NODE_DEBUG=release-it:*` as documented here: https://github.com/release-it/release-it?tab=readme-ov-file#troubleshooting--debugging

Any chance this could still be worked in?

lib/util.js
```diff
@@ -1,5 +1,6 @@
-import fs from 'node:fs';
+import fs, { close, openSync, statSync, utimesSync } from 'node:fs';
```
webpro, line 1: Maybe good for consistency to either always use `fs` or the named exports?

test/plugins.js
```diff
@@ -1,6 +1,7 @@
 import { join } from 'node:path';
+import sh from 'node:child_process';
```
webpro, line 2: Let's use the same name for this default import everywhere.

lib/util.js
```diff
@@ -1,5 +1,6 @@
...
 import { EOL } from 'node:os';
+import { execSync } from 'node:child_process';
```
webpro, line 3: This import doesn't seem to be used (found this by running `npm run lint`)

test/util/sh.js
```diff
@@ -0,0 +1,26 @@
+import childProcess from 'node:child_process';
+
+const getCommandAndArgs = input => {
+  //"git checkout master".split(' ').slice(1)
+  if (!input || (input && typeof input !== 'string'))
+    throw new Error(`No command given to getCommandAndArgs. Please check input and/or type`);
+
+  const command = input.split(' ').reverse().pop();
```
webpro, line 8: Splitting the command by space (same for `args` below) is maybe good enough if we use it only in our own tests. Is this something shelljs used an actual parser for?

webpro (commented):
Thanks @phun-ky! Much appreciated, would be great to get rid of shelljs by using Node.js built-ins. I do have a few comments but hopefully we can work those out.

lib/shell.js
phun-ky: From what I see, the config here is only for `lib/shell.js` right? and that hindered `shelljs` to pipe its `stdout` to `process.stdout`. But in `lib/shell.js`, you already expect `stdout`. So either I am not following the logic here, or I need to ignore `stdout` when `!debug.enabled`

lib/util.js
phun-ky: Yup

lib/shell.js
phun-ky: which brings up the question, you use `execa` as well. wouldn't it be wise to replace that as well (in another PR?)

test/plugins.js
phun-ky: Fixed

lib/util.js
phun-ky: Fixed

test/util/sh.js
phun-ky: `shelljs` does alot of file i/o just for the `execSync` parth that is overkill imho, and use `execArgs` via a `paramsFile`: https://github.com/shelljs/shelljs/blob/8a960da8383666024bfdb2eff709a4fdf66333ee/src/exec.js#L100

might need a separate test yeah. will add

phun-ky: Doh, looked at it wrongly. writing test for a test util? *facepalm*. Have amended the code

lib/shell.js
phun-ky: Ah, ok. This will apply to exec-stuff _outside_ of `lib/shell.js` from what I can tell, since `execStringCommand` only displays output on log level verbose, which is debug. Right?

phun-ky: I think I have the solution. Stay tuned

webpro: That sounds about right indeed. I think the behavior can be seen most clearly when comparing output of running the tests in `main` with running them in this branch, as in `main` there's almost no output other than what the test runner Ava logs. Same when actually running release-it as a user (I sometimes do this during development using `npm link`).

phun-ky:
@webpro should be fixed now! 🇳🇴 delivers <3

webpro:
Thanks a bunch, @phun-ky! 🇳🇴 LFG 🚀

Think I'm going to merge and make it a new major release (it's quite the change after all!), and pre-release under the `alpha` or `next` tag first.

phun-ky:
Glad to help! BTW, I see there's some possibility for more housekeeping changes that I can gladly add if interested. Will ofc make issues for discussion first

webpro:
> Glad to help! BTW, I see there's some possibility for more housekeeping changes that I can gladly add if interested. Will ofc make issues for discussion first

Awesome. Feel free to discuss and/or open PRs, they're most welcome :) Btw: https://bsky.app/profile/webpro.nl/post/3lihsjotnvg2x

ghiscoding:
@webpro is it a good assumption to think that this is the PR that made it to be v19.0 major? Because I wasn't sure what was the breaking change(s), it's typically useful to see directly in the release before deciding to upgrade right away or wait a little more... in other words, more info in the release is always good. Thanks for this awesome project 🚀

webpro:
Good question! This is a fairly large change as there might be differences in the way shell command execution happens between ShellJS and the Node.js built-in `child_process` and if any release-it config would be relying on specific shelljs behavior then that could be a breaking change. I'm also thinking about potential differences across Windows and *nix here.

Semver is about expectations and I think a minor bump could lead to unexpected surprises here. Perhaps it's mostly my own ignorance that drove this decision.

So currently there's a pre-release (`next`), v19.0.0 isn't released yet. After some mileage we'll publish it. 

https://github.com/release-it/release-it/blob/main/CHANGELOG.md will contain the major updates.

ghiscoding:
Yeah I know it was tagged as alpha, though I was curious if this PR was the one contributing to the major bump and it looks like it is. Adding to the changelog when ready for the final release seems like a good plan. Thanks for confirming. Cheers

PR #1202, opened 2025-02-19
chore: remove `lodash` as a direct dependency

phun-ky:
2n round of housekeeping. Removing `lodash` as a direct dependency. 

**Things to do after this:**

* Find a better way to get default options (getting rid of `@nodeutils/defaults-deep` that was added in this PR)
* Strive for better readability (having a _hard_ time wrapping my head around the current code format/structure)
* Consider replacing `ava` with the native node test runner. It's faster, and `ava` uses `lodash` either as direct dependency or 2nd level dependency)

**Next steps:**

* Refactor (it's a bit messy)
* Use TS (would've helped greatly when making changes tbh, not a requirement but can help)
* Tweak ESLint-config


phun-ky:
@webpro ping

lib/util.js
```diff
@@ -132,7 +144,144 @@ const tryStatFile = filePath => {
...
+const pick = (object, keys) => {
+  return keys.reduce((obj, key) => {
+    if (object && Object.prototype.hasOwnProperty.call(object, key)) {
+      obj[key] = object[key];
+    }
+    return obj;
+  }, {});
+};
+
+const merge = (target, ...sources) => {
```
webpro, line 237: Removing a fairly large dependency like Lodash is great, but from a maintainer's perspective inlining such non-trivial functions isn't great. What do you think about using small focused dependencies? Or something else that does not require me to maintain this myself?

package.json
```diff
@@ -69,6 +69,7 @@
     "lint": "eslint lib test",
     "format": "prettier --write eslint.config.mjs \"{lib,test}/**/*.js\"",
     "docs": "remark README.md 'docs/**/*.md' '.github/*.md' -o",
+    "test2": "NODE_DEBUG=release-it* ava --no-worker-threads && installed-check --ignore ava --ignore nock",
```
webpro, line 72: Please either remove this or give the script a more descriptive name

webpro (commented):
Thanks for the PR, Alexander! I like it (and the plans too), but to be honest I don't fancy maintaining the utilities myself :)

phun-ky:
@webpro will look through the comments. 

I dislike being dependent on dependencies that:

* Are outdated
* Are not required (vanilla alternative)
* Are too small to _not_ handle yourself

When you have the utilities yourself, you do not required to update them in the same pace as a dependency, and you know that they are locked to match your code. That is my two cents at least. YMMV

package.json
```diff
@@ -79,6 +80,7 @@
   "license": "MIT",
   "dependencies": {
     "@iarna/toml": "2.2.5",
+    "@nodeutils/defaults-deep": "1.1.0",
```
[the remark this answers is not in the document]
phun-ky: @webpro yep, that was the only way to keep the current default options as is without rewriting the whole options part. You have several deps that use lodash as transitive dependencies: sinon, issue-parser, ava etc.

lib/util.js
phun-ky: in my experience and view, if a dependency is too small, do not use it as a dep/create a dep for it. Please see the pr comment: https://github.com/release-it/release-it/pull/1202#issuecomment-2684236494

But, we are different people with different views, if you think that splitting up in minor/small dependencies is the way to go, that is your call :)

package.json
phun-ky: @webpro that was an oversight on my behalf. will fix!

webpro:
While I agree with all of your points, perhaps the first question is: what exactly is "too small"? To me, built-ins and two-liners are basically no-brainers, happy to use those. But imho functions like `templateInterpolate` and `merge` are not that small. The other question is: what consequences does it have? They're not only handling internal code, they would also take in external user templates and configuration objects (and also through (un)known external plugins). There are no sources or references, no tests, but those do represent more code/risks for me to manage.

To make it explicit, these 4 functions I would reconsider:

- `templateInterpolate`
- `isPlainObject`
- `get`
- `merge`

At the very least have links to sources, references, tests and/or attribution.

Would you agree? Is there any chance you'd be willing to work this in? I haven't actually looked into locations or packages for this, maybe you have ideas?

phun-ky:
When it comes to "too small", it boils down to what your app is trying to be. for example, my @phun-ky/speccer app is trying to have a zero `dependency` state, so I have no production deps. 

In other cases, I find that if the function/code is more than the IDE/Editor viewport, refactor it to multiple files. If the code you are splitting out can be useful for others, publish it as a package. If the code you have written is over the viewport size (approx 50-80 lines of code) and it has already been published as a package that _only_ covers that part of code, sure, use that. (I still have `leftpad` in my mind..)

I've refactored away `get`, and use a dep for `isObjectLoose/isObjectStrict` (aka `isPlainObject/isObject`. (shameless plug, since I remembered I actually released those features in my package yesterday :S :P )

The most time spent with getting rid of `lodash` was to find a suitable drop in replacement for the `_.template` function. That took blood, sweat, tears, chatgpt and trying to reverse engineer what `lodash` was doing without _all of the bloat_ they used. I see that you already have tests for `format`, so from that I guess that you have the input under control (?).

I've added tests for `merge`.

webpro:
> When it comes to "too small", it boils down to what your app is trying to be. for example, my @phun-ky/speccer app is trying to have a zero `dependency` state, so I have no production deps.
> 
> In other cases, I find that if the function/code is more than the IDE/Editor viewport, refactor it to multiple files. If the code you are splitting out can be useful for others, publish it as a package. If the code you have written is over the viewport size (approx 50-80 lines of code) and it has already been published as a package that _only_ covers that part of code, sure, use that. (I still have `leftpad` in my mind..)

Sorry, not sure what to make of this.

> I've refactored away `get`, and use a dep for `isObjectLoose/isObjectStrict` (aka `isPlainObject/isObject`. (shameless plug, since I remembered I actually released those features in my package yesterday :S :P )
> 
> The most time spent with getting rid of `lodash` was to find a suitable drop in replacement for the `_.template` function. That took blood, sweat, tears, chatgpt and trying to reverse engineer what `lodash` was doing without _all of the bloat_ they used.

You should know I really appreciate the efforts and I feel bad for saying this, but I hope you understand this does not make me prefer this over the original Lodash functions. I prefer battle-tested functions with some mileage.

If `_.template` doesn't have a light-weight drop-in replacement available then I guess it can be dealt with separately and get replaced with something else in some next major update of release-it.

> I see that you already have tests for `format`, so from that I guess that you have the input under control (?).
>
> I've added tests for `merge`.

Unfortunately, in some cases both the template and data are defined elsewhere. Third-party plugins and users of release-it can define their own templates rendered with context/data merged from release-it defaults, plugin defaults and user configuration.

Overall, at this point I'm not convinced it's a good idea to replace all of Lodash with alternatives. I'd say we can do it in steps and, for now, leave in the 4 Lodash methods I mentioned in my previous comment? That will still really help future work and improvements we can do.

phun-ky:
image

Will switch deps for this

webpro:
> > Sorry, not sure what to make of this.
> 
> I was just answering your question: `what exactly is "too small"?`. I can take the time to elaborate more on this, if my answer did not suffice :)

Guess we were talking past each other a bit. I meant "too small" as some kind of lower limit when choosing whether or not to _consume_ an external dependency (i.e. not when to _publish_).

> > You should know I really appreciate the efforts and I feel bad for saying this, but I hope
> > you understand this does not make me prefer this over the original Lodash functions. I prefer
> > battle-tested functions with some mileage.
> 
> There are several reasons why using `lodash` is bad practice, and I think you and I are on the same page here.

💯

> I can understand you reasoning, but I am a bit perplexed. If every open source maintainer had that standpoint, there would be no room for small players, and that is not [sustainable](https://phun-ky.net/2023/10/09/making-open-source-sustainable-challenges-and-solutions).

I meant this in the context of going from a Lodash method to an alternative. In general, it's normal for projects to "work their way up" and gain levels of trust (this could form and shape in various ways and could take some time).

> Just to be clear, we're not at an impasse here, just want you to have that awareness in mind :)
> 
> When it comes to "battle hardened", for `isObjectStrict` and `isObjectLoose`, I can't think about more ways to test them:
> 
> https://github.com/phun-ky/typeof/blob/main/src/__tests__/typeof.ts#L140 https://github.com/phun-ky/typeof/blob/main/src/__tests__/typeof.ts#L85

In my mind, "battle" is more about "out there in the wild/trenches" as opposed to local unit tests. Including some time to "harden" by receiving feedback and bug fixes, etc. Certainly not only unit tests.

> I would be happy to reconsider some of the functions, of course! We in the Open Source community help each other out you know ;)

💯

> * `_.templateInterpolate` is a no-brainer, but for future replacement you need to add more tests to cover the user custom input in `format` to cover your bases, and perhaps reconsider the templating setup.
> * `_.isPlainObject` is a carbon copy of [`isObjectStrict`](https://github.com/phun-ky/typeof/blob/main/src/main.ts#L104).
> * `_.get`: You did not fully utilize all of the features that this function has, there was mainly usage that gets the first level property of the given object, and 4-5 usages for chained property gets: `repo.project` and `repo.host`. And it seemed odd to use a dependency for that tbh.
> * `_.merge`, yeah, this one I can use the `lodash` version of, since it is not a fully _logical_ merge. This function can in the future be replaced by something else when you also get rid of the `_.defaultsDeep` function
> 
> 🧑‍🏭

Thanks! I'll take some time tomorrow to look at the code and see where we are with this.

> Note
> 
> Lodash has not been tested with Node versions over v12

Good point, we certainly agree it's a good idea to replace Lodash with alternatives.

lib/plugin/GitRelease.js
```diff
@@ -21,12 +20,13 @@ class GitRelease extends GitBase {
       'changelog',
       'commit'
     ]);
-    return _.defaults(baseOptions, gitOptions);
+    const a = Object.assign({}, gitOptions, baseOptions);
+    return a;
```
webpro, line 24: Can be returned directly without assigning to `a` first

lib/plugin/factory.js
```diff
@@ -54,23 +54,23 @@ export const getPluginName = pluginName => {
 export let getPlugins = async (config, container) => {
   const context = config.getContext();
   const disabledPlugins = [];
-
-  const enabledExternalPlugins = await _.reduce(
-    context.plugins,
-    async (result, pluginConfig, pluginName) => {
```
webpro, line 60: Thanks for this refactor! Was quite horrible 🙈

lib/util.js
```diff
@@ -1,70 +1,103 @@
...
+      autoEscape: false,
+      useWith: true,
+      tags: ['${', '}'],
+      parse: { interpolate: '' },
+      rmWhitespace: false,
+      autoTrim: false
+    });
+    return eta.renderString(template, context);
   } catch (error) {
+    const log = new Log();
```
webpro, line 68: Why is this moved here, would also expect this at module scope?

lib/config.js
```diff
@@ -82,15 +85,16 @@ class Config {
...
     );
   }
-
   getContext(path) {
-    const context = _.merge({}, this.options, this.contextOptions);
-    return path ? _.get(context, path) : context;
+    let context = merge({}, this.options, this.contextOptions);
+    if (!path) return context;
+
+    return context[path];
```
webpro, line 92: See other comment about `_.get`

webpro (commented):
Thanks Alexander, did a bit of a deeper dive and left some comments. Hope you can agree!

lib/plugin/Plugin.js
```diff
@@ -40,12 +40,30 @@ class Plugin {
...
+      if (!context) {
+        return undefined;
+      }
+      context = context[prop];
+      if (context === undefined) {
+        return undefined;
+      }
+    }
+
+    return context;
```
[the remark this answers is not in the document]
phun-ky: ok. will change

lib/config.js
phun-ky: will fix

lib/util.js
```diff
@@ -1,70 +1,103 @@
...
   try {
-    return _.template(template)(context);
+    const eta = new Eta({
+      autoEscape: false,
+      useWith: true,
+      tags: ['${', '}'],
+      parse: { interpolate: '' },
+      rmWhitespace: false,
+      autoTrim: false
+    });
```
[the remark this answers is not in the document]
phun-ky: will fix

lib/util.js
phun-ky: with the current setup, i get `ReferenceError: Cannot access 'Log' before initialization` when moving it to module scope. Since it is only used there, I see no reason to move it away from where it is used

lib/plugin/Plugin.js
```diff
@@ -40,12 +41,13 @@ class Plugin {
...
-    const context = _.merge({}, this.options, this.context);
-    return path ? _.get(context, path) : context;
+    const context = merge({}, this.options, this.context);
+
+    return path ? get(context, path) : context;
   }
 
   setContext(context) {
-    _.merge(this.context, context);
+    this.context = merge(this.context, context);
```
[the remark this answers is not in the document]
phun-ky: will fix

lib/util.js
```diff
@@ -1,70 +1,102 @@
...
+export const parseGitUrl = remoteUrl => {
   if (!remoteUrl) return { host: null, owner: null, project: null, protocol: null, remote: null, repository: null };
   const normalizedUrl = (remoteUrl || '')
     .replace(/^[A-Z]:\\\\/, 'file://') // Assume file protocol for Windows drive letters
     .replace(/^\//, 'file://') // Assume file protocol if only /path is given
     .replace(/\\+/g, '/'); // Replace forward with backslashes
   const parsedUrl = gitUrlParse(normalizedUrl);
   const { resource: host, name: project, protocol, href: remote } = parsedUrl;
-  const owner = protocol === 'file' ? _.last(parsedUrl.owner.split('/')) : parsedUrl.owner; // Fix owner for file protocol
+  const owner = protocol === 'file' ? parsedUrl.owner.split('/').slice(-1)[0] : parsedUrl.owner; // Fix owner for file protocol
```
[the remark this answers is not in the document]
phun-ky: will fix

webpro:
Thanks a bunch for bearing with me, Alexander! Very glad we got this in.

phun-ky:
open source collaboration ftw!

webpro:
:rocket: _This pull request is included in v19.0.0-next.1. See [Release 19.0.0-next.1](https://github.com/release-it/release-it/releases/tag/19.0.0-next.1) for release notes._

PR #1207, opened 2025-03-13
fix: don't use stdout for logs

mdvorak:
Don't use console.log for logging, as it outputs to stdout instead of stderr.

When used from scripts, it is impossible to use e. g. `release-it --release-version --increment major` with conventional-changelog, because that outputs
```
VERSION=WARNING The recommended bump is "minor", but is overridden with "major".
1.0.0
```
instead of two streams, and version only in stdout.

I've checked that `--release-version` and `--changelog` arguments already use (correctly) `console.log` directly, and I've replaced log for cli commands.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -1,10 +1,7 @@
 import { readJSON } from './util.js';
-import Log from './log.js';
 import runTasks from './index.js';
 
 const pkg = readJSON(new URL('../package.json', import.meta.url));
-
-const log = new Log();
 
 const helpText = `Release It! v${pkg.version}
 
@@ -27,10 +24,10 @@ const helpText = `Release It! v${pkg.version}
 For more details, please see https://github.com/release-it/release-it`;
 
 /** @internal */
-export const version = () => log.log(`v${pkg.version}`);
+export const version = () => console.log(`v${pkg.version}`);
 
 /** @internal */
-export const help = () => log.log(helpText);
+export const help = () => console.log(helpText);
 
 export default async options => {
   if (options.version) {
--- a/lib/log.js
+++ b/lib/log.js
@@ -16,7 +16,7 @@ class Logger {
   }
 
   log(...args) {
-    console.log(...args);
+    console.error(...args);
   }
 
   error(...args) {


mdvorak:
If you would consider this change too much, less obtrusive variant would be replace it only for warn (and possibly info), and leave log method outputting to stdout.

webpro:
> Don't use console.log for logging, as it outputs to stdout instead of stderr.

I guess it depends on exactly what output we're talking about. Not ALL logging must go to `stderr`. What specifically is problematic or incorrect?

> When used from scripts, it is impossible to use e. g. `release-it --release-version --increment major` with conventional-changelog, because that outputs
> 
> ```
> VERSION=WARNING The recommended bump is "minor", but is overridden with "major".
> 1.0.0
> ```
> 
> instead of two streams, and version only in stdout.

I would agree that the first line with the warning should not go to `stdout`, if that's what you mean?

> I've checked that `--release-version` and `--changelog` arguments already use (correctly) `console.log` directly, and I've replaced log for cli commands.

There's a separate `--ci` mode which is automatically enabled in CI environments.

The default mode will ask feedback from the user and print changelog, etc. - this could go to `stdout` just fine I'd say.

> If you would consider this change too much, less obtrusive variant would be replace it only for warn (and possibly info), and leave log method outputting to stdout.

It's a bit much indeed, but the good thing is that your timing is great as we're heading towards a new v19 which could introduce potentially breaking changes like this. Please consider this my initial review, I'll give it a closer look when I can.

mdvorak:
OK, so proposed variant would be:

Log warnings and info to stderr, but `log()` method should still print to stdout, to follow console own behavior. I'll verify usage in the code.

This should work fine in CI, since commands that are useful in shell scripts should not use `log()`, even in nested code (I'll also verify this).

mdvorak:
Pushed updated version, plus I've added `process.exit(1)` for `--release-version` and `--changelog`, if the command failed. Otherwise it prints either empty string, or worse, `"undefined"`.

lib/index.js
```diff
@@ -95,13 +100,14 @@ const runTasks = async (opts, di) => {
...
 
-    if (!version) {
-      log.obtrusive(`No new version to release`);
-    }
-
     if (isReleaseVersion) {
-      console.log(version);
-      process.exit(0);
+      if (version) {
+        console.log(version);
```
webpro, line 105: While I like the idea and result of this all, here's a bit of a nitpick: why not use the `Log` class here anymore? In my mind, the `Log` class for logging, to abstract away `stdout` vs `stderr` and no need to use `console.*` directly.

webpro (approved):

lib/index.js
mdvorak: Well, it is not logging, its direct output of the command. But it does not really matter, I did the changes incrementally, and with final version, its not needed to modify this file anymore.

webpro:
:rocket: _This pull request is included in v19.0.0-next.2. See [Release 19.0.0-next.2](https://github.com/release-it/release-it/releases/tag/19.0.0-next.2) for release notes._

Issue #934, opened 2022-09-09
Idea: Shareable configurations


elkevinwolf:
It would be nice if `release-it` can support shareable configurations, like `eslint` and `prettier` does.

An idea on how to implement this is to replace `cosmiconfig` with [`c12`](https://github.com/unjs/c12), which offers out-of-the-box:

- support for `extends` syntax
- support for typescript configurations, as well as json, cjs and rc files

benediktberger:
Have my upvote, that would be a nice feature indeed. We have many projects doing basically the same thing on release, would appreciate a shared config.

webpro:
I'm happy to review a PR for this, as long as it keeps supporting the existing options (also see https://github.com/release-it/release-it/blob/master/lib/config.js).

webpro:
Closing this due to inactivity. Feel free to re-open or submit a PR.

PR #1211, opened 2025-03-25
Use URL to parse extended configuration string

webpro:
Parse incoming extended config string using `new URL()`.

Coming from https://github.com/release-it/release-it/pull/1181#discussion_r2003592360

What do you think, @juancarlosjr97? This is what I had in mind, but happy to hear your thoughts here? The tests still pass, but maybe I'm missing certain edge or use cases?

Btw we've also moved to [mentoss](https://mentoss.dev) for mocking `fetch` requests which I think is quite nice :)

PR #1212, opened 2025-03-28
feat: use c12 to replace cosmiconfig

aa900031:
Follow that issue #934, we can use c12 to replace cosmiconfig.

### Changes

1. The `options`, and `localConfig` is not sync anymore in `Config`:

Because the loadConfig provided by c12 is processed asynchronously, we need to change the way obtains `options` and `localConfig` in class `Config`. 
For the above reasons, an additional `resolved` variable is provided so that we can await it and ensure that `options` and `localConfig` have been obtained correctly.

2. Adjust the way to get config in tests, to support asynchronous


webpro:
Thanks @aa900031! 

#1181 was just merged to add support for `extend` configuration.

What's the difference? Can we end up with one or the other, but not both?

aa900031:
Hi @webpro,

I think the biggest difference is that c12 supports nested extends and multiple extends, for example:

In my-project, extends two configs

```json
// my-project/.release-it.json
{
    "extends": [
        "github:owner/repo",
        "github:foo/bar",
    ]
}
```

These two configs also extended from other repos

```toml
# github/owner/repo/.release-it.toml
extends = [
  "github:other/repo"
]
[git]
commitMessage = "Release version ${version}"
```

```yaml
# github/other/repo/.release-it.yaml
gitlab:
  release: true
```

Another key point in the example above is that extend doesn't have to be in a JSON file.

So switching to c12 will make the functionality more complete and increase flexibility.

package.json
```diff
@@ -94,11 +94,13 @@
     "issue-parser": "7.0.1",
     "lodash.get": "4.4.2",
     "lodash.merge": "4.6.2",
+    "make-synchronized": "0.6.0",
```
webpro, line 97: This isn't used?

webpro:
Thanks for the explanation, that makes sense.

The async config loading adds a lot of code overall, but I guess sooner or later this had to happen anyway.

@juancarlosjr97 What do you think?

package.json
aa900031: Sorry my bad, that is try to make load config to synchronized, but that have some problems

I'm already drop that at d3e7fbb

aa900031:
@webpro There tests have some problems about mock fetch response, here is the PR https://github.com/humanwhocodes/mentoss/pull/92

docs/configuration.md
```diff
@@ -91,9 +91,11 @@ Release-it has [plenty of options][2]. See the following tables for plugin confi
...
 You can extend a configuration from a remote source using the `extends` option. The following formats are supported:
 
-- `github:owner/repo`
-- `github:owner/repo#tag`
-- `github:owner/repo:file#tag`
+- `github:owner/repo`: Get the config from the default branch (main) in the repo at Github.
+- `github:owner/repo#tag`: Get the config from the specified tag in the repo at Github.
+- `github:owner/repo:subdir#tag`: Get the config from the specified tag in the repo sub dir at Github.
+
+And support other schema, either `gitlab:`, `bitbucket:`, or `https:`.
```
juancarlosjr97, line 98: If it is HTTPS, then the file should be public and not behind authorization. However, GitHub, GitLab, and BitBucket will work as it is getting pulled with the token, right?

juancarlosjr97:
Everything looks good, a very nice improvement, only one question.

juancarlosjr97:
Also, this will make this change https://github.com/release-it/release-it/pull/1211/files not longer needed @webpro I believe

docs/configuration.md
aa900031: Yes you right, if want to get private repo or private content from https, it could use the environment `GIGET_AUTH` to set it. The value will be set in `Authorization: Bearer ...` header

Refs to [giget docs](https://github.com/unjs/giget?tab=readme-ov-file#providing-token-for-private-repositories)

juancarlosjr97: Cool! Then I am super happy! I did not know we were working with giget, but I can see it is a transitive dependency of c12, so awesome :)

webpro:
> Also, this will make this change https://github.com/release-it/release-it/pull/1211/files not longer needed @webpro I believe

True! Thanks for chiming in, very glad to have your opinion on this one.

webpro:
About the `Config` constructor. Currently it automatically starts an async loader, while `constructor` itself is sync, which might be a bit unclear that we need `config.resolve`:

```ts
const config = new Config();
await config.resolved;
```

Is this a well-known pattern to add async logic to the constructor? If not, how about this:

```ts
const config = new Config();
await config.resolve(); // or await config.init();
```

So that the async part isn't automatically triggered from the `constructor`, but it would be made more explicit by having to call `resolve()` or `init()` on the instance.

For me personally this latter pattern feels more natural, but happy to hear your thoughts.

aa900031:
@webpro I think that make more sense as you say. I like this pattern to use `await config.init()`, so I'll refactor that later to fit this pattern.

aa900031:
Hi @webpro , [mentoss](https://github.com/humanwhocodes/mentoss/releases/tag/mentoss-v0.9.2) was released include fix, and I'm already upgraded it. Please check it again, thanks

webpro:
Thank you @aa900031! Also for the fix in Mentoss 🔥

webpro:
:rocket: _This pull request is included in v19.0.0-next.5. See [Release 19.0.0-next.5](https://github.com/release-it/release-it/releases/tag/19.0.0-next.5) for release notes._

PR #1214, opened 2025-04-04
fix(json-schema): change `addUntrackedFiles` to boolean

KyleRoss:
The JSON Schema for the `git` plugin had `addUntrackedFiles` set as a string type when it should be a boolean. This leads to a warning when setting this option in `.release-it.json`. This only applies when a JSON file is used for configuration.

Screenshot 2025-04-04 at 12 46 11 PM

### DIFF
--- a/schema/git.json
+++ b/schema/git.json
@@ -38,8 +38,8 @@
       "default": ""
     },
     "addUntrackedFiles": {
-      "type": "string",
-      "default": ""
+      "type": "boolean",
+      "default": false
     },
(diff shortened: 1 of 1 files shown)


webpro:
Thanks Kyle!

webpro:
:rocket: _This pull request is included in v19.0.0-next.4. See [Release 19.0.0-next.4](https://github.com/release-it/release-it/releases/tag/19.0.0-next.4) for release notes._

PR #1215, opened 2025-04-07
fix: parse boolean values from command line

mdvorak:
Don't use "false" string as true value.

For example `release-it --git.commit=false` was parsed as `true`, because string `"false"` is non-empty.

### DIFF
--- a/bin/release-it.js
+++ b/bin/release-it.js
@@ -1,74 +1,11 @@
 #!/usr/bin/env node
 
 import updater from 'update-notifier';
-import parseArgs from 'yargs-parser';
 import release from '../lib/cli.js';
 import { readJSON } from '../lib/util.js';
+import { parseCliArguments } from '../lib/args.js';
 
 const pkg = readJSON(new URL('../package.json', import.meta.url));
-
-const aliases = {
-  c: 'config',
-  d: 'dry-run',
-  h: 'help',
-  i: 'increment',
-  v: 'version',
-  V: 'verbose'
-};
-
-const booleanOptions = [
-  'dry-run',
-  'ci',
-  'git',
-  'npm',
-  'github',
-  'gitlab',
-  'git.addUntrackedFiles',
-  'git.requireCleanWorkingDir',
-  'git.requireUpstream',
-  'git.requireCommits',
-  'git.requireCommitsFail',
-  'git.commit',
-  'git.tag',
-  'git.push',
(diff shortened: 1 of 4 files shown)


bin/release-it.js
```diff
@@ -16,9 +16,41 @@ const aliases = {
   V: 'verbose'
 };
 
+const booleanOptions = [
+  'dry-run',
+  'ci',
+  'git.requireCleanWorkingDir',
```
webpro, line 22:
```suggestion
  'git.addUntrackedFiles',
  'git.requireCleanWorkingDir',
```

bin/release-it.js
```diff
@@ -16,9 +16,41 @@ const aliases = {
   V: 'verbose'
 };
 
+const booleanOptions = [
+  'dry-run',
+  'ci',
```
webpro, line 21:
```suggestion
  'ci',
  'git',
  'npm',
  'github',
  'gitlab',
```

webpro (commented):
Thanks Michal

The idea is that you can negate using `--no-` e.g. `--no-git.commit` (https://www.npmjs.com/package/yargs-parser#boolean-negation)

It's cool if we'd be able to use `=false` (and `=0`?) to negate as well, but it is not documented: https://www.npmjs.com/package/yargs-parser - any chance you could squeeze in a test somewhere?

mdvorak:
> Thanks Michal
> 
> The idea is that you can negate using `--no-` e.g. `--no-git.commit` (https://www.npmjs.com/package/yargs-parser#boolean-negation)
> 
> It's cool if we'd be able to use `=false` (and `=0`?) to negate as well, but it is not documented: https://www.npmjs.com/package/yargs-parser - any chance you could squeeze in a test somewhere?

Actually, both are supported by yargs, where --no flags being an extra feature (can be disabled). When using in scripts, its often more readable to do `--foo=$FOO_ENABLED` then building `FOO_ARG` variable. Yargs is sadly lacking documentation in some regards.

About unit test, thats good idea.

mdvorak:
I've refactored bin to args.js, so it is testable. Testing whole bin js is problematic (but can be done).

webpro:
:rocket: _This pull request is included in v19.0.0-next.4. See [Release 19.0.0-next.4](https://github.com/release-it/release-it/releases/tag/19.0.0-next.4) for release notes._

PR #1221, opened 2025-04-29
fix: remove lodash

jmfrancois:
Looking at https://www.npmjs.com/package/lodash.get

image

I have spot that issue coming from a deprecation warning message by using release-it with pnpm.


webpro:
Thanks, @jmfrancois! Does this support paths like `key[0].into.deep[1]`? Would be great if there'd be test or two to confirm. Any chance you could add some?

webpro:
Closing this in favor of #1231.

PR #1224, opened 2025-05-25
feat: add gitlab.repoId option

microHoffman:
This PR introduces a new release-it option `gitlab.repoId`.

## Summary
I've ran into an issue during when a `GitLab.isCollaborator` function was called. The request to 

```
RELEASE-IT:GITLAB 66: {
  url: 'https://mygitlab.com/api/v4/projects/namespace%2Fprojectname/members/all/22',
  method: 'GET'
}
RELEASE-IT:GITLAB 66: Error: 404 Not Found
ERROR User name is not a collaborator for ... .
```

returned 404, which release-it misinterpreted as an issue that a user is not a collaborator (which is not an issue here). I've found out the error is thrown because our current GitLab API instance does not support using namespaced paths in place of IDs. This can be the case for multiple reasons, e.g. different [server config](https://wiki.dongfg.com/base/gitlab-api-404.html) settings.

While I think the usually correct resolution is to fix the server settings, sometimes it's not really possible to adjust the server config ourselves (for whatever reason), which is why I needed to implement this PR to get release-it working for me. After I've specified `gitlab.repoId`, the request to Gitlab API finally went through fine and the release was made:)

Let me know if you think it's useful to merge, just wanted to create a PR in case it is:)

Thanks for good project!

### DIFF
--- a/config/release-it.json
+++ b/config/release-it.json
@@ -69,6 +69,7 @@
     "useGenericPackageRepositoryForAssets": false,
     "genericPackageRepositoryName": "release-it",
     "origin": null,
-    "skipChecks": false
+    "skipChecks": false,
+    "repoId": null
   }
 }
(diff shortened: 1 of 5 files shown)


webpro (commented):
Happy to merge, but I do have a nitpick (re. docs)

webpro:
:rocket: _This pull request is included in v21.0.0. See [Release 21.0.0](https://github.com/release-it/release-it/releases/tag/21.0.0) for release notes._

PR #1229, opened 2025-05-30
Fix GitLab secure configuration default value

AlejandroRM-DEV:
### Problem
The documentation states that `gitlab.secure` has a default value of `false`:

```
gitlab.secure: Flag to disable server certificate verification (default: `false`) 
```

However, the actual default value is `null`, which causes the following comparison to evaluate incorrectly:

```javascript
const needsCustomAgent = Boolean(secure === false || certificateAuthorityFile);
```

When `secure` is `null`, the condition `secure === false` returns `false`, leading to unexpected behavior in the custom agent logic.

### Solution
- Updated the default value of `secure` to `false` in both the GitLab configuration and schema
- This ensures consistency between the documented behavior and actual implementation

### Additional Changes
- Updated documentation to clarify that GitLab tokens require `api` and `self_rotate` scopes

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,7 @@ GitLab projects can have releases attached to Git tags, containing release notes
 releases][32]:
 
 - Configure `gitlab.release: true`
-- Obtain a [personal access token][33] (release-it only needs the "api" scope).
+- Obtain a [personal access token][33] (release-it needs the `api` and `self_rotate` scopes).
 - Make sure the token is [available as an environment variable][34].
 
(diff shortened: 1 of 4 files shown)


webpro:
Thanks for the catch & the fix, Alejandro! 👌

webpro:
:rocket: _This pull request is included in v19.0.3. See [Release 19.0.3](https://github.com/release-it/release-it/releases/tag/19.0.3) for release notes._

PR #1231, opened 2025-06-23
Replace lodash.get with custom get() function and add tests

AlejandroRM-DEV:
## Description

Captura de pantalla 2025-06-23 a la(s) 4 59 05 p m

This PR removes the usage of the deprecated `lodash.get` function and replaces it with a native JavaScript implementation that:

* Supports deep access to nested properties via string paths with dot notation, bracket notation for arrays, and quoted keys (e.g., `'a.b[0]["c.d"]'`).
* Uses optional chaining internally to safely traverse objects without throwing errors.
* Allows specifying a default fallback value when the requested path does not exist.

## Testing

* Tests cover:

  * Simple nested property access.
  * Array indices.
  * Keys containing dots or special characters.
  * Non-existent paths with default values.
  * Empty or invalid paths.
* All tests pass successfully.

Captura de pantalla 2025-06-23 a la(s) 4 52 26 p m

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,7 @@ GitLab projects can have releases attached to Git tags, containing release notes
 releases][32]:
 
 - Configure `gitlab.release: true`
-- Obtain a [personal access token][33] (release-it only needs the "api" scope).
+- Obtain a [personal access token][33] (release-it needs the `api` and `self_rotate` scopes).
 - Make sure the token is [available as an environment variable][34].
 
 → See [GitLab Releases][35] for more details.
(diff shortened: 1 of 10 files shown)


stylder:
Hi! 👋 Just wanted to say thanks for the work on this PR.

This feature seems quite useful, especially for those of us using release-it in automated workflows. Since the implementation looks solid, is well-documented, and checks are passing — would it be possible to get this merged soon?

I believe it could benefit a lot of users facing similar scenarios. If there's anything needed to move it forward, happy to help however I can.

Thanks again for maintaining such a great project!

stylder:
@webpro

Issue #1258, opened 2025-10-28
Feature Request:  Option to Limit or Disable Changelog Logging in Console


Napster2709:
**Problem Description**

Currently, the changelog is written directly to the console log during the release process. For projects with a long changelog, this results in the console being flooded with the entire changelog, making it difficult to focus on the relevant information.

**Proposed Solution**

Introduce a configuration option to control the changelog logging behavior in the console. The following options could be provided:

1.  Log only the changes for the next release: Display only the changelog entries relevant to the upcoming release.
2.  Disable changelog logging: Provide an option to completely suppress the changelog output in the console.

j-ulrich:
It seems the requested feature was released with [version 20.1.0](https://github.com/release-it/release-it/releases/tag/20.1.0). So I guess this issue can be closed.

Napster2709:
indeed

PR #1274, opened 2026-01-21
feat: add --quiet flag to hide release previews

Yeom-JinHo:
#1258 closed

## Description

This PR introduces a `quiet` flag that suppresses the preview blocks printed during the release process.

Previously, release-it would print the full changelog (and, via the same `Logger.preview()` call, the changeset and release notes) during the preview phase, which could flood the console for projects with long changelogs or noisy diffs.
With this change, users can suppress that preview output while keeping the `--changelog` behavior unchanged.

## Changes

- **New top-level option** `quiet` (boolean, default `false`) with matching CLI flag `--quiet`.
  The name mirrors the conventional meaning in other CLIs (e.g., `git --quiet`, `npm --quiet`) and keeps the config surface flat rather than introducing a new nested namespace.
- **Logger-level guard**: `Logger.preview()` now short-circuits when `isQuiet` is set. This covers all three preview callsites — changelog (`lib/index.js`), changeset (`lib/plugin/git/Git.js`), and release notes (`lib/plugin/GitRelease.js`) — so the flag name matches its scope.
- **User-visible notice**: When `quiet` is active, a single `Preview output hidden (--quiet).` line is logged right after the "Let's release ..." banner, so suppression is never silent.
- **CLI parser**: `quiet` added to `booleanOptions` in `lib/args.js` so `--quiet` / `--no-quiet` parse as proper booleans.
- **Schema update**: `schema/release-it.json` adds a top-level `quiet: boolean` (default `false`) with a description covering all three previews. `additionalProperties: false` preserved.
- **Type update**: `types/config.d.ts` adds `quiet?: boolean`.
- **Default config**: `config/release-it.json` sets `quiet: false`.
- **Docs**: `docs/changelog.md` section updated to cover the new flag and scope.
- **Tests**:
  - `test/log.js` — unit tests assert that all three preview titles are suppressed when `isQuiet: true`, and still rendered when `false`.
  - `test/tasks.js` — integration tests cover default behavior, `quiet: false`, and `quiet: true` (including the info notice).

## Motivation

- **Reduce console noise**: Long changelogs and diffs make it hard to focus on the important release output and can push the confirmation prompt off-screen.
- **Keep existing workflows**: Users who rely on `--changelog` see no behavior changes — that code path exits before the preview block.
- **Honest naming**: A flag called `quiet` should affect all preview output, not just one block. A Logger-level guard keeps the name and the scope aligned.

## Breaking Changes

None for released versions.
Default remains `false`, preserving current output behavior.
`--changelog` mode is untouched.

> Note: this PR replaces an earlier unreleased `logs.preview.changelog` nested option (from previous commits on this branch) with the simpler top-level `quiet` flag. No user-facing breakage since that shape never shipped.

### DIFF
--- a/config/release-it.json
+++ b/config/release-it.json
@@ -1,4 +1,5 @@
 {
+  "quiet": false,
   "hooks": {},
   "git": {
     "changelog": "git log --pretty=format:\"* %s (%h)\" ${from}...${to}",
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -6,6 +6,22 @@ commits since the latest tag.
 The [default command][1] is based on `git log ...`. This setting (`git.changelog`) can be overridden. Make sure any of
 these commands output the changelog to `stdout`.
 
+## Hide changelog preview
+
+For projects with very long changelogs, you can hide the changelog preview during the release preview to avoid flooding
+the console. This only affects the preview output during the normal release flow, not the `--changelog` mode which prints
+the changelog and exits.
+
+```json
+{
+  "logs": {
+    "preview": {
+      "changelog": "hide"
+    }
+  }
+}
+```
+
 - [GitHub and GitLab Releases][2]
 - [auto-changelog][3]
 - [Conventional Changelog][4]
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -6,20 +6,22 @@ commits since the latest tag.
 The [default command][1] is based on `git log ...`. This setting (`git.changelog`) can be overridden. Make sure any of
 these commands output the changelog to `stdout`.
 
-## Hide changelog preview
+## Hide release preview output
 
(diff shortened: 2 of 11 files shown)


Yeom-JinHo:
@webpro 
Would you mind taking a moment to review this PR  when you have a chance?

webpro:
Thanks for the PR! Tbh it feels a bit much for what's being asked for. How about something like a `--quiet` flag (and `quiet: true` option) and something along the lines of:

```ts
  preview({ title, text }) {                                      
    if (this.isQuiet) return;
    // ...existing body                                                                                                                                                                                                                            
  }
```

Could something like this cover the needs as well?

Yeom-JinHo:
@webpro 
 Thanks — reworked as suggested. --quiet / quiet: true with an early return in Logger.preview().

webpro:

PR #1275, opened 2026-01-22
refactor: Continuing my previous refactor work to remove lodash

phun-ky:
This is not meant to be a "replace anything with my own packages", but rather a quest to get rid of `lodash` and find smaller, preferably vanilla, alternatives.

I could not find any decent alternatives, and `@node-utils/defaults-deep` (which I added back in the days) still had `lodash` as a dependency. This is now removed.

Next up is `issue-parser`.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -1,7 +1,7 @@
 import util from 'node:util';
 import assert from 'node:assert';
 import { isCI } from 'ci-info';
-import defaultsDeep from '@nodeutils/defaults-deep';
+import defaultsDeep from '@phun-ky/defaults-deep';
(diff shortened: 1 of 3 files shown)


webpro:
Thanks for the pull request! The underlying goal here, removing the full `lodash` package pulled in by `@nodeutils/defaults-deep`, was completed in https://github.com/release-it/release-it/pull/1297 and released in v20.0.1.

release-it now uses `defu`, with an explicit compatibility layer and regression tests for the previous array-replacement behavior. The current dependency graph no longer contains the full `lodash` package; the small standalone `lodash.merge` package remains in use.

Since this branch now conflicts with and is superseded by that work, I’m closing this PR.

PR #1289, opened 2026-04-01
fix: allow false as npm config value in types

ahippler:
`false` is a valid value to disable publishing to npm, see https://github.com/release-it/release-it/blob/main/docs/npm.md?plain=1#L7

The current types do not allow `false` as value.

### DIFF
--- a/types/config.d.ts
+++ b/types/config.d.ts
@@ -99,7 +99,7 @@ export interface Config {
 
     /** @default 10 */
     timeout?: number;
-  };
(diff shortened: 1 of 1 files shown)


webpro:
Thanks Andreas!

webpro:
:rocket: _This pull request is included in v20.0.1. See [Release 20.0.1](https://github.com/release-it/release-it/releases/tag/20.0.1) for release notes._

PR #1299, opened 2026-04-27
feat: loosen version specifications of dependencies

SSW-SCIENTIFIC:
Current version specifications on `package.json`, is very tight except some packages like `defu`, which has `^` in its version. This tight version specifications ensure release-it works on well-tested dependency versions, on the other hand, pin vulnerable versions by release-it if users use release-it via `package.json`'s `devDependencies`.

This is trade-off between stablity and security, however, by default users can depends on `package-lock.json`'s dependency versions, so I think it is better to loosen `package.json`'s dependency versions. This allow that users can use release-it with stable, well-tested dependencies by default, and can update minor or patch version of dependencies with Dependabot or `npm audit fix` if necessary.


panuva-kp:
@webpro This would be great. As there is again undici vuln on a specific version (this time 7.28 fixes it), and likely there will be another one in a month. Currently all projects utilizing `release-it` report high-classified vulns on transitive deps which you cannot fix with `npm audit fix` but only with overrides of each vulnerable dependency on the main project.

panuva-kp (commented):
Probably `npm install` 1st to sync the package-lock.json with the caret changes could be a good idea?

webpro:
Loosening also means a compromised version of a (transitive) dependency can enter the scene. Or just annoying bugs. There is a reason I switched to fixed versions.

However, switching to (pnpm and) `minimumReleaseAge` should help us out here. I'll see what I can do shortly. Closing this one as it's too intrusive for my likings. Thanks for bringing it to my attention though!

PR #1303, opened 2026-05-29
fix: ensure release body is an empty string instead of null

driiftkiing:
Error: 
```
ERROR 422 (Invalid request.

For 'properties/body', nil is not a string. 
```

Reproduce:
Create release with custom changelog, but with a empty changelog.


webpro:
Thank you!

webpro:

Issue #1300, opened 2026-05-19
Race conditions?


Mopsgamer:
```diff
$ bun run release-it --increment=minor --ci
bun prerelease
+ ✔ bun prerelease
- ERROR Working dir must be clean.
Please stage and commit your changes.
Documentation: https://git.io/release-it-git
error: "release-it" exited with code 1
error: script "release:minor" exited with code 1
Error: Process completed with exit code 1.
```

Mopsgamer:
Have enabled `--bun` and it is working. Weird. I've decided it's a new bug in Node or V8 microtasks around `then` method lowering. Can't find any public reports/evidence. 

<img width="220" height="164" alt="Image" src="https://github.com/user-attachments/assets/43cd4a90-2702-4ff9-90db-63ae1cc1de45" />

webpro:
Bun isn't officially supported.

From the report I can't figure out whether the working dir is actually clean? Or some input or output that could provide some hints either.

You could [troubleshoot](https://github.com/release-it/release-it#troubleshooting--debugging) and analyze what's going on in the process. Going to close this one, feel free to let me know if there's something actionable for me.

webpro:
Maybe https://github.com/release-it/release-it/issues/687 / https://github.com/release-it/release-it/pull/688 are related.

Mopsgamer:
@webpro yes, it's a duplicate. I know Bun isn't officially supported, but this is just a shell command; I'm still using npm for the actual publishing. Bun is only being used to run scripts faster.

I haven't been able to get a proper reproduction though.
[This PR comment](https://github.com/release-it/release-it/pull/688#issuecomment-664380031) suggests using `git update-index -q --refresh && git diff-index --quiet HEAD --`, but I can't tell if that was ever actually tried.

I suspect the core issue might be what's described in [this Stack Overflow comment](https://stackoverflow.com/questions/2657935/checking-for-a-dirty-index-or-untracked-files-with-git/2659808#comment40313224_2659808).

Running bun prerelease can definitely touch files, which is likely triggering the dirty index.

PR #1304, opened 2026-05-30
fix: dirty tree detection

Mopsgamer:
Ref: #687, #1300

Replaces `git diff --quiet HEAD` with `git update-index -q --refresh` followed by `git diff-index --quiet HEAD --`.

This should fix intermittent issues with dirty tree detection.

While this is difficult to test reliably, `git diff` refreshes the index internally but can return early before the index is fully and synchronously updated. Running `git update-index -q --refresh` is the safest way to ensure the index is fully updated before making a comparison. Additionally, `git diff` is a porcelain command designed for human use; switching to the plumbing command `git diff-index --quiet HEAD --` provides a more stable and reliable check for automation.

**Performance Note:**
Tested using `hyperfine`, the execution speeds of the two diff commands themselves are practically identical. Explicitly running `update-index` does double the execution time (from ~67ms to ~120ms on my system), but the absolute overhead is negligible and well worth the guarantee of a reliable dirty check.

### DIFF
--- a/lib/plugin/git/Git.js
+++ b/lib/plugin/git/Git.js
@@ -124,10 +124,12 @@ class Git extends GitBase {
   }
 
   isWorkingDirClean() {
-    return this.exec('git diff --quiet HEAD', { options }).then(
-      () => true,
-      () => false
-    );
+    return this.exec('git update-index -q --refresh', { options })
+      .then(() => this.exec('git diff-index --quiet HEAD --', { options }))
+      .then(
+        () => true,
+        () => false
+      );
   }
 
   async getUpstreamArgs(pushRepo) {
--- a/test/git.js
+++ b/test/git.js
@@ -313,8 +313,8 @@ describe('git', () => {
     await gitClient.tag();
     await gitClient.rollbackOnce();
 
-    assert.equal(exec.mock.calls[11].arguments[0], 'git tag --delete v1.2.4');
-    assert.equal(exec.mock.calls[12].arguments[0], 'git reset --hard HEAD~1');
+    assert.equal(exec.mock.calls[12].arguments[0], 'git tag --delete v1.2.4');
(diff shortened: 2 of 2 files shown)


webpro:
Thanks for digging into this and measuring the overhead. I went back through #687 and #688, then tested this change against main.

#688 considered these same commands. We chose `git diff --quiet HEAD` because porcelain `git diff` enables [`diff.autoRefreshIndex`](https://git-scm.com/docs/git-diff) by default. Git refreshes stat-only entries and checks file content before returning. I found no evidence that it can return before its refresh completes, and I could not reproduce the linked failure.

The existing command had 0 false-dirty results in 1,000 touched-file runs. Across both implementations, a release-it fixture had 0 failures in 80 Node/Bun runs. The changes to `test/git.js` only adjust mock call indexes; they do not show main failing where this branch passes.

The proposed sequence also changes behavior in two cases:

- If `.git/index.lock` exists, `git update-index --refresh` exits 128 and a clean repository is reported as dirty. The existing command reports it as clean.
- `git diff-index` does not honor the porcelain-only [`diff.ignoreSubmodules`](https://git-scm.com/docs/git-diff) setting. Repositories using that setting can therefore be reported as dirty.

I did find a different edge case. If a staged change is undone only in the worktree (`HEAD=A`, `index=B`, `worktree=A`), `git diff --quiet HEAD` reports clean. This PR catches it, but it is not the #1300 failure and has no test here.

A safer fix would check both sides explicitly with `git diff --quiet --` and `git diff --cached --quiet HEAD --`, plus a test for this exact state.

Without a reproduction for the linked failure, I don’t want to replace the existing check with one that adds these regressions. I’m closing this PR. Thanks for putting the proposal together; it helped uncover the staged/unstaged edge case.

PR #1305, opened 2026-06-01
feat: add requireVersionFail option to exit 0 when no new version

### LINKED ISSUE #1277: Add option to exit with code 0 when no new version to release
**Problem**
To strictly comply with conventional-commits and bump version only for breaking changes or non-hidden commit types, the `bumpStrict` [option](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-conventionalcommits#specific-options) must be enabled. In this scenario, when version is not bumped (which is ok), release-it returns exit code 1 and a warning (which is not ok):
```
npm run -s version && echo 0 || echo 1
...
WARNING No new version to release
1
```
https://github.com/release-it/release-it/blob/61e6939575d72105730e5efff652989ce020d6b5/lib/index.js#L104-L112
This behavior causes the CI pipeline to fail instead of terminating gracefully and skipping the next steps.
The current solution is to parse changelog entries in pipeline, like [this](https://github.com/conventional-changelog/conventional-changelog/issues/399#issuecomment-3419604041) or this:
```
if release-it --ci --changelog | grep -wqE '^### (Features|Bug Fixes)'; then ...
```
which is not reliable, because changelog entries can be changed or reconfigured. 

**Solution**
Add `requireVersionFail` option, which will keep current behaviour when enabled (default) or will return exit code 0 without warning when disabled. This will help to improve the reliability of CI pipelines.

viraatdas:
Fixes #1277

## Problem
When release-it determines there is no new version to release (for example under conventional-commits with `bumpStrict`), it logs a warning and exits 1, which fails CI pipelines that legitimately have nothing to release.

## Fix
Add a top-level `requireVersionFail` option (default `true`, preserving current behavior). When set to `false` (via `--no-requireVersionFail`), release-it exits 0 in that case instead of 1. The name mirrors the existing `git.requireCommitsFail` convention, and the option is threaded through args, CLI help, schema, types, defaults, and docs.

## Tests
Added a regression test asserting exit 0 when `requireVersionFail` is false and no new version is available, plus a guard that the default still exits 1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,9 @@ Alternatively, a plugin can be used to override this (e.g. to manage a `VERSION`
 
 Add the `--release-version` flag to print the **next** version without releasing anything.
 
+By default, `--release-version` exits with code `1` when no new version can be determined. Set
+`requireVersionFail: false` or pass `--no-requireVersionFail` to exit with code `0` without a warning instead.
+
 ## Git
 
 Git projects are supported well by release-it, automating the tasks to stage, commit, tag and push releases to any Git
--- a/config/release-it.json
+++ b/config/release-it.json
@@ -1,5 +1,6 @@
 {
   "quiet": false,
+  "requireVersionFail": true,
   "hooks": {},
   "git": {
     "changelog": "git log --pretty=format:\"* %s (%h)\" ${from}...${to}",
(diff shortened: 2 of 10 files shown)


webpro:
Thanks for working on this. The linked issue #1277 has been resolved on main by 56c70c3, which makes --release-version exit successfully and silently when no version is available and includes regression coverage for the external-plugin path. This PR now conflicts with that implementation, so I am closing it as superseded.

PR #1306, opened 2026-06-04
Validate missing config argument

### LINKED ISSUE #1196: Error messages due to mistakes with command line options are not clear
**Bug Report**

**Title:**
Unclear Error Message: "The `paths[0]` argument must be of type string. Received type boolean (true)"

**Description:**
When running the `release-it` CLI with the following malformed command-line arguments:

```sh
yarn release-it --ci --no-git.commit --no-npm.publish --config --git.pushRepo=gitlab .release-it.ts --pre-release
```

the error message is:

```
The "paths[0]" argument must be of type string. Received type boolean (true)
```

This error message is confusing and does not clearly indicate which part of the input caused the issue. The problem lies in the improper usage of `--config`, which must be immediately followed by the config file name, but another argument (`--git.pushRepo=gitlab`) was placed in between.

**Steps to Reproduce:**
1. Run the following command in a terminal:
   ```sh
   yarn release-it --ci --no-git.commit --no-npm.publish --config --git.pushRepo=gitlab .release-it.ts --pre-release
   ```
2. Observe the error message.

**Expected Behavior:**
The error message should clearly point out the incorrect usage of the `--config` argument and suggest the proper syntax. For example:

```
Invalid argument: "--config" must be immediately followed by the configuration file name.
```

**Actual Behavior:**
The error message does not provide sufficient context or guidance to identify the incorrect argument:

```
The "paths[0]" argument must be of type string. Received type boolean (true)
```

**Environment:**
- `release-it` version: 18.1.1
- Node.js version: 22.13.0
- Operating System: mac OS 13.5.2

**Additional Context:**
This issue occurs because the argument parser does not enforce that `--config` is followed directly by a valid file name. When another argument (e.g., `--git.pushRepo=gitlab`) intervenes, it leads to an invalid state that triggers a cryptic error message.

**Proposed Solution:**
- Improve the argument parser to validate the usage of `--config` and ensure it is immediately followed by a valid file name.
- Update the error message to clearly indicate that `--config` was used incorrectly and provide guidance on the proper syntax.
- Enhance documentation to warn against separating `--config` and its associated value.

**Workaround:**
Ensure that `--config` is immediately followed by the config file name. For example:

```sh
yarn release-it --ci --no-git.commit --no-npm.publish --config .release-it.ts --git.pushRepo=gitlab --pre-release
```

hyjcxy:
Fixes #1196.

This adds a CLI argument validation step for `--config`/`-c` so release-it reports a clear error when the config option is followed by another flag instead of a configuration file name.

Validation run:
- `node --env-file=.env.test --test test/args.js`
- `npm run lint`

I also tried `npm test`, but the existing extended configuration fetch tests failed locally with `Invalid configuration file at .release-it`, and the full run later stalled in unrelated integration tests.

### DIFF
--- a/lib/args.js
+++ b/lib/args.js
@@ -1,4 +1,4 @@
-import parseArgs from 'yargs-parser';
+import { parseArgs } from 'node:util';
 
 const aliases = {
   c: 'config',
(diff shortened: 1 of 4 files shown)


webpro:
We should probably start using e.g. `node:util.parseArgs` (and probably start with loose `strict: false`). Any chance you'd be up for that?

hyjcxy:
Updated this to use `node:util.parseArgs` with `strict: false`.

I kept a small compatibility layer for the current parser behavior:
- dotted options are expanded back into nested objects
- known boolean option values such as `--dry-run=false` are normalized to booleans
- short aliases and positional increment behavior are preserved
- loose unknown string options like `--git.pushRepo origin` are still handled

Validation run:
- `node --env-file=.env.test --test test/args.js`
- `npm run lint`

webpro:
My point was basically that if we use `parseArgs` there is no need to validate that e.g. `--config` is a string arg.

webpro:
Closing this with caf77c928bfefa48339e8246261ffb8e41d70a4c

PR #1312, opened 2026-06-29
Fix npm publish workspace config resolution

### LINKED ISSUE #1308: regression: npm.publish with --workspaces=false ignores monorepo root npm config
After upgrading to `release-it@20.2.0`, publishing scoped workspace packages no longer uses the registry configured in the monorepo root npm config.

The regression from v19.1 appears related to commit https://github.com/release-it/release-it/commit/14a4907c56fabaa970e0fde3221cb700fc52556c, which added `--workspaces=false` to `npm publish`:

This fixes the previous npm workspace null / matches error, but it changes npm config resolution when the command is executed from an individual workspace package directory.

In our case, release-it runs inside each package directory. With --workspaces=false, npm no longer applies the monorepo root .npmrc, so the scoped package registry is lost and npm falls back to https://registry.npmjs.org/.

### Environment

release-it: 20.2.0
npm: 11.12.1
node: 24.15.0
OS: Linux
Monorepo tool: npm workspaces + Turbo
Package manager config: root .npmrc
Monorepo Setup
Root .npmrc:
```
@custom:registry=http://verdaccio:4873
```

### Repro steps

A direct dry-run reproduces the same behavior outside release-it:

```
cd packages/telemetry
npm publish . --tag rc --workspaces=false --dry-run

npm notice Publishing to https://registry.npmjs.org/ with tag rc and default access (dry-run)
```

### Expected Behavior
release-it should preserve the registry behavior from previous versions when publishing a workspace package.

### Possible Fix
Maybe release-it could make the `--workspaces=false` behavior configurable for npm publish, or preserve the root workspace npm config when running publish commands from workspace package directories.

For example:
```
npm: {
  workspaces: false
}
```

But the current hardcoded --workspaces=false makes it difficult to keep root npm config as the source of truth in npm workspace monorepos.

cyphercodes:
## Summary
- Stop forcing `--workspaces=false` on `npm publish` / `npm stage publish` so npm can resolve workspace-root `.npmrc` scoped registry config.
- Keep the existing `npm version --workspaces=false` behavior unchanged.
- Update npm/task tests to cover publish commands without the forced workspace flag.

Fixes #1308

## Test plan
- `npm ci`
- `node --env-file=.env.test --test test/npm.js`
- `node --env-file=.env.test --test test/tasks.js`
- `npm run lint`
- `git diff --check`
- `npm test`
- Manual dry-run repro: `npm publish . --tag rc --dry-run` uses the workspace root scoped registry, while adding `--workspaces=false` falls back to npmjs.org.

### DIFF
--- a/lib/plugin/npm/npm.js
+++ b/lib/plugin/npm/npm.js
@@ -277,16 +277,9 @@ class npm extends Plugin {
       this.log.warn('Skip publish: package is private.');
       return false;
     }
-    const args = [
-      publishPath,
-      '--tag',
-      tag,
-      publishPackageManager === 'npm' && '--workspaces=false',
-      ...otpArgs,
-      dryRunArg,
-      registryArg,
-      ...fixArgs(publishArgs)
-    ].filter(Boolean);
+    const args = [publishPath, '--tag', tag, ...otpArgs, dryRunArg, registryArg, ...fixArgs(publishArgs)].filter(
+      Boolean
+    );
     const publishCommand = stage ? ['stage', 'publish'] : ['publish'];
     const isInteractive = !stage && (!this.config.isCI || Boolean(this.config.isPromptOnlyVersion));
     return this.exec([publishPackageManager, ...publishCommand, ...args], {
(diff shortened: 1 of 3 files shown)


webpro:
Thanks Rayan, this should be fixed indeed.

webpro:
:rocket: _This pull request is included in v21.0.0. See [Release 21.0.0](https://github.com/release-it/release-it/releases/tag/21.0.0) for release notes._

PR #1314, opened 2026-07-06
docs: migrate to Starlight

trueberryless:
#### Description

This PR uses the Markdown files in this repo to create a documentation website built with [Starlight](https://starlight.astro.build/).
It also adapts some part so they suite better on the website and refactors some docs to follow the [Diátaxis](https://diataxis.fr/) framework. All the information that were in `/docs` before are migrated to the docs on the website. No information got lost.

Sorry about the huge diff, but I think splitting this into multiple PRs is even harder because everything is strongly related.

Please follow [the "Deploy from a custom action" guide](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-with-a-custom-github-actions-workflow) before merging 🦖 

<details><summary>Some screenshots</summary>
image
image
image
image

</details>


webpro:
Thank you for putting so much thought and substantial work into this, Felix!

This would add another surface for the project to maintain. I would prefer to keep the documentation simply as Markdown in the repository.

That said, there as some very useful ideas such as "Your first release" and the "How release-it works" explanation so those ideas are worth considering adoption. If you're interested, smaller focused improvements to the existing Markdown documentation based on those ideas would be welcome. But please don’t feel obliged to extract them. I appreciate the work you’ve already done here.

This is a project-direction and maintenance decision, not a reflection on the effort you put into this PR. I also don't want to ask you to dismantle such a large PR when we won't adopt its main element, so I'm going to close it.

trueberryless:
Thanks for the kind words!

I understand the decision because of the additional maintenance. I'll keep the branch, just in case you want to extract some parts of it. I myself don't expect to find the time to extract parts, but we'll see 👍

Issue #42, opened 2015-11-19
Add an option to use the repository from the package.json


hitsthings:
That would be the canonical repo, so would make sense (at least in my case) to use it as "src.pushRepo".

webpro:
The default (for `git push`) is whatever is configured to be the push origin in the local repo (see `git remote -v`). Also, I think by default the `repository.url` in `package.json` should match this (not the other way around).

Apparently, you have a different setup, so you can override `src.pushRepo` any way you like.

Please let me know if I'm misunderstanding you. Feel free to re-open, but I'm closing it for now.

hitsthings:
I normally have that setup too, and most people will. But the git configuration is defined by the local dev and the codebase itself has no input into how the local setup looks. Conversely, the package.json is defined by the codebase and can point to the known canonical repo no matter what the local setup looks like. If I'm publishing to NPM, then I want the tags in the canonical repo.


Take this situation.

1. git clone (origin URL) repo
2. cd repo
3. (Do some work but don't finish)
4. (Quick! There is a production bug to fix! Let's create a second working directory to deal with it)
5. cd ..
6. git clone ./repo repo2 # local clone is faster
7. cd repo2
8. (Fix bug quick!)
9. release-it patch

Now I've pushed the tag to ./repo instead of (origin URL) which is incorrect. If release-it used the package.json value I would have been fine.

Does this make sense?

You also mentioned that I can set src.pushRepo to whatever I want, but hardcoding the value twice in the codebase is not DRY and could cause issues if we move the repo later. What I've done so far is set up a script that reads the value from package.json and then programmatically calls release-it with a generated config.

I raised this improvement so it could be easier for others in the future. But if you see things differently that's cool. Cheers!

webpro:
Regarding the extra local clone. In the situation you describe, I would *stash* my work, go to the branch, fix, release, change back to branch I was working in, apply stash.

And, imho the need to change this in two locations (if at all? YAGNI?) doesn't weigh up against creating that script. DRY is great, but not at all costs.

hitsthings:
Thanks Lars, I think I just realized that `release-it` is not for me. Not because you're not implementing this (like I said, I'd already solved it), but because the value I was getting out of using it was already minimal, and it sounds like you're generally not interested in supporting more flexibility in workflows.

To address your two points (just for the sake of offering another perspective - I no longer have a vested interested in this):

[...]

PR #1324, opened 2026-08-06
Prevent duplicate comments on resolved issues and PRs

### LINKED ISSUE #1: fail on push
Hi,

First thanks for this tool which is very useful to maintain my micro libs.

However I found that it always fail on the git push step for me,
I think it should do an explicit "git push origin master" instead of "git push" because that second command is likely to fail for various reason.

For instance if you have other branches (e.g. `gh-pages` and out of sync, the push will try to push them all and may fails).

Also one of my other library was using the `"git://"` syntax which doesn't seem to work with this library (or I'm wrong?).

Thanks,

krisu-pl:
## Problem

When `github.comments.submit` is enabled, the same item can receive the release comment multiple times:

- **Pull requests** — commit SHAs from the changelog are batched into multiple search queries (GitHub's search API caps queries at 256 encoded characters). A pull request whose commits span more than one batch is returned by each query, and each occurrence was commented on.
- **Issues** — an issue referenced by multiple changelog entries (e.g. two commits both saying `fixes #1`) was parsed once per reference, and commented on once per reference.

So a release with enough commits could leave multiple identical " This issue has been resolved in ..." comments on the same PR or issue.

CleanShot 2026-08-05 at 16 40 57@2x

## Fix

Deduplicate both lists at the source:

- `commentOnResolvedItems` – merged pull request numbers from all search results go through a `Set` before being commented on
- `getResolvedIssuesFromChangelog` – resolved issue numbers are deduplicated the same way

## Tests

- Added a new test: a changelog with enough commits to produce two search queries, where both queries return PR `#42`, the test checks if only one comment has been added.
- New test case for `getResolvedIssuesFromChangelog` with the same issue number referenced on two changelog lines, but should get a comment only once

### DIFF
--- a/lib/plugin/github/GitHub.js
+++ b/lib/plugin/github/GitHub.js
@@ -459,7 +459,12 @@ class GitHub extends Release {
 
     const shas = getCommitsFromChangelog(changelog);
     const searchResults = await Promise.all(searchQueries(this.client, owner, repo, shas));
-    const mergedPullRequests = searchResults.flatMap(items => items.map(item => ({ type: 'pr', number: item.number })));
+    // A pull request whose commits span more than one search query is returned by each of them,
+    // so the numbers must be deduplicated to comment only once per pull request.
+    const mergedPullRequests = [...new Set(searchResults.flat().map(item => item.number))].map(number => ({
+      type: 'pr',
+      number
+    }));
 
     const hostURL = 'https://' + (this.options.host || host);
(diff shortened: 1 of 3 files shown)


webpro:
Thanks Chris! Needed this myself too 👌

webpro:
:rocket: _This pull request is included in v21.0.2. See [Release 21.0.2](https://github.com/release-it/release-it/releases/tag/21.0.2) for release notes._

PR #1327, opened 2026-08-07
fix: strip control characters from changelog before display

### LINKED ISSUE #1326: Changelog preview isn't stripped of terminal escape sequences
## Summary

release-it generates a changelog from `git log` commit subjects and displays it both on every default run, as part of the standard interactive release preview, and via the documented `--changelog` flag (`lib/index.js`, `lib/log.js`). The value is written to the terminal with no filtering of control characters. Unlike a package name or a git branch/tag name, a commit message has no character restrictions. Since running `release-it` is exactly what a maintainer does to prepare a release, and the changelog preview is shown by default, any commit in the repository's history containing an embedded escape sequence in its message is written to the terminal and interpreted live during a routine release.

## POC

(available upon request)

## Impact

I validated this by creating a commit with a message containing a raw OSC title-set escape sequence, then running release-it's changelog display (`--changelog`) in that repository inside a tmux session and capturing the literal byte stream written to the pty via `tmux pipe-pane` (which taps the raw bytes before tmux's own screen-buffer interpretation) and inspecting it as a hex dump. The captured stream contained the exact injected sequence, unmodified, embedded in the commit subject text, confirming it reaches the terminal without any stripping. I confirmed via the project's own test suite that this same changelog value is shown by default on every run, not only via the `--changelog` flag used for a cleaner reproduction. I used a benign title-set marker specifically to confirm the sequence reaches output unstripped, rather than a more aggressive payload; the specific range of what an injected OSC/CSI sequence can do depends on the victim's terminal emulator and was not enumerated further.

## Suggested Fix

Strip control characters, particularly the ESC byte, from the changelog text before it's written to the terminal. I'd suggest fixing this once, where the changelog value is obtained from the plugin chain in `lib/index.js`, since both the `--changelog` output and the default interactive preview consume the same value.

carfeii:
Fixes #1326

The changelog, generated from `git log` commit subjects, is shown by default on every run as part of the interactive release preview, and via `--changelog`, with no filtering of control characters. Unlike a package name or a git ref, a commit message has no character restrictions, so a crafted commit message could inject terminal escape sequences into release-it's own output during a routine release.

## Change

Adds `sanitizeForDisplay` (`lib/util.js`), which filters out characters with a code point below 32 or equal to 127, other than newline (preserved since a changelog legitimately spans multiple lines). Applied once, where the changelog value is obtained from the plugin chain in `lib/index.js`, so it covers both the `--changelog` output and the default interactive preview.

## Testing

- Reproduced the issue with a commit whose message contained a raw OSC title-set escape sequence, confirmed via a `tmux pipe-pane` capture (hex-dumped, not just visually inspected) that the raw escape bytes reached the terminal embedded in the commit subject text.
- Re-ran the same reproduction against this fix; the injected control bytes are gone, only harmless printable text remains.
- Ran the relevant parts of the test suite (`test/tasks.js`, `test/log.js`, `test/utils.js`, `test/util/*.js`, 52 tests total): all pass, including the tests specifically asserting the default changelog preview still renders correctly.
- Note on environment: the available Node here was v18.19.1, below the `engines.node: "^22.21.0 || >=24.0.0"` requirement, which caused a hard `EBADENGINE` install failure; used a Node v24.4.1 binary for installing dependencies and running tests, no project files depend on this.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -4,7 +4,7 @@ import Config from './config.js';
 import Shell from './shell.js';
 import Prompt from './prompt.js';
 import Spinner from './spinner.js';
-import { reduceUntil, parseVersion, castArray } from './util.js';
+import { reduceUntil, parseVersion, castArray, sanitizeForDisplay } from './util.js';
 
 const runTasks = async (opts, di) => {
   let container = {};
@@ -60,7 +60,7 @@ const runTasks = async (opts, di) => {
 
     const name = await reduceUntil(plugins, plugin => plugin.getName());
     const latestVersion = (await reduceUntil(plugins, plugin => plugin.getLatestVersion())) || '0.0.0';
-    const changelog = await reduceUntil(plugins, plugin => plugin.getChangelog(latestVersion));
+    const changelog = sanitizeForDisplay(await reduceUntil(plugins, plugin => plugin.getChangelog(latestVersion)));
 
     if (isChangelog) {
       if (changelog) {
--- a/lib/util.js
+++ b/lib/util.js
@@ -173,6 +173,17 @@ export const getNpmEnv = () => {
   return env;
 };
 
+// Strip control characters (e.g. terminal escape sequences) from text
+// that originates from user-controlled sources such as git commit
+// messages, which have no character restrictions, before it is written
+// to the terminal. Newline is preserved since changelogs legitimately
+// span multiple lines.
+export const sanitizeForDisplay = string =>
+  string &&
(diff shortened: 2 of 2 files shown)


webpro:
A (bad) solution for a problem we don't have:

- How do such escape codes even enter a changelog?
- Better solution would be: https://nodejs.org/api/util.html#utilstripvtcontrolcharactersstr
- No tests at all. At first sight this looks like it would strip more than intended.

Issue #598, opened 2020-02-08
TypeScript ?


joris-:
👋  Hi ! 

What about use TypeScript or add type defs ? 
It could be a good thing for creating Plugins for example.

webpro:
Yes, especially for the Plugin stuff that would be great. Any chance you could start this thing? What would be the best way to add typings to this project?

joris-:
Yes, I've start to working on a type defs file.

We can add typings by 3 ways : 
- Re-write project in TS
- Add types to https://github.com/DefinitelyTyped/DefinitelyTyped
- Or add types to this project (<= I choose this)

webpro:
I don't have much experience with TS other than writing TS from scratch (or join existing TS codebase). I don't really like the idea of rewriting the whole project now, so let's go for it. Thanks! We can always add these to DefinitelyTyped later, I guess.

joris-:
So I've spend times to get type defs but I'm not satisfied. Maybe someone will get a better result.

PR #1330, opened 2026-08-13
Add type declarations for Plugin and related types

j-ulrich:
Adds type declarations for the `Plugin` and related classes.

See also #598


types/plugin.d.ts
```diff
@@ -0,0 +1,73 @@
...
+  verbosityLevel: number;
+  isDebug: boolean;
+  isQuiet: boolean;
+  isCI: boolean;
+  isPromptOnlyVersion: boolean;
+  isReleaseVersion: boolean;
+  isChangelog: boolean;
+
+  options: Options;
+  localConfig: any;
```
j-ulrich, line 25: I was not sure what `localConfig` is. Is this bascially the same as `options`?!
Or is it considered an internal property anyway? In that case we should probably rather remove it.

types/index.d.ts
```diff
@@ -1,3 +1,4 @@
 declare module 'release-it';
 
 export type { Config } from './config.d';
+export type { default as Plugin } from './plugin.d';
```
webpro, line 4: [P1] Export Plugin as a value

`import { Plugin } from 'release-it'; class MyPlugin extends Plugin {}` fails with TS1362 because `Plugin` is exported using `export type`. Extending the class is the documented plugin entry point. Export the class as a value, matching `lib/index.js`.

types/plugin.d.ts
```diff
@@ -0,0 +1,73 @@
...
+  prompt: Prompt;
+  shell: Shell;
+}
+
+export default class Plugin {
+  constructor({ namespace, options, container }: { namespace: string; options: Options; container: Container });
+
+  init(): void;
+  getName(): string;
+  getLatestVersion(): string;
```
webpro, line 49: [P2] Match lifecycle arguments and asynchronous results

Valid overrides such as `async getLatestVersion() { return '1.2.3'; }` and `bump(version: string) {}` both fail with TS2416. The runner awaits getters and supplies arguments to `getChangelog`, the increment methods, and `bump`. Declare those arguments and allow promises and the supported no-result values for getters; the base implementations themselves return `undefined`.

types/plugin.d.ts
```diff
@@ -0,0 +1,73 @@
...
+  getIncrement(): string;
+  getIncrementedVersionCI(): string;
+  getIncrementedVersion(): string;
+  beforeBump(): void;
+  bump(): void;
+  beforeRelease(): void;
+  release(): void;
+  afterRelease(): void;
+
+  options: Options;
```
webpro, line 60: [P2] Type plugin options separately from global configuration

`this.options` contains the result of `getInitialOptions`, which defaults to the plugin's own configuration. It is not the global `Config`. A plugin declaring `options: { filename: string }` fails with TS2416, although that configuration works at runtime. Give plugin options their own extensible type or generic, while keeping the global configuration type on `this.config.options`.

types/plugin.d.ts
```diff
@@ -0,0 +1,73 @@
...
+  config: Config;
+  log: Logger;
+  shell: Shell;
+  spinner: Spinner;
+  prompt: Prompt;
+  debug: DebugLogger;
+
+  getContext(path: string): any;
```
webpro, line 68: [P2] Include the documented context and step helpers

`this.setContext(...)` and `this.step(...)` fail because neither method is declared. `this.getContext()` and `this.config.getContext()` also fail because their path parameters are required. These calls are supported at runtime and documented for plugin authors. Add the missing methods and make both context paths optional.

types/plugin.d.ts
```diff
@@ -0,0 +1,73 @@
...
+  exec(command: string, { options, context }: { options: Record<string, any>; context: Context }): Promise<any>;
```
webpro, line 69: [P2] Preserve exec defaults and template context

Both `this.exec('git status')` and the documented `this.exec('git log', { options: { write: false } })` fail because the declaration requires the second argument and both its fields. A supplied template context also incorrectly requires `log`, `shell`, `spinner`, and `prompt`. Match the runtime defaults, allow a data object for context, and make the underlying `Shell.exec` options optional too.

types/prompt.d.ts
```diff
@@ -0,0 +1,41 @@
...
+export type UnknownPromptConfig = { type: string } & MessageProvider & Partial<ChoicesProvider> & {
+    default?: any;
+    transformer?: (value: any) => any;
+    validate?: (value: any) => boolean | string | Promise<boolean | string>;
+  };
+
+export type PromptConfig = ConfirmPromptConfig | InputPromptConfig | ListPromptConfig | UnknownPromptConfig;
+
+export default interface Prompt {
+  register(pluginPrompts: PromptConfig[], namespace?: string): void;
```
webpro, line 26: [P2] Register prompts by name

Both registration declarations require `PromptConfig[]`, but the implementation merges an object keyed by prompt name and later looks up that name. `registerPrompts({ publish: { type: 'confirm', message: () => 'Publish?' } })` fails with TS2353. An array is accepted by the declaration but registers numeric keys, so showing `publish` throws. Use `Record<string, PromptConfig>` in both declarations.

webpro (changes requested):
Thanks for adding these declarations! Several signatures reject supported plugin implementations, and the type-only export prevents extending Plugin. Details inline.

types/index.d.ts
j-ulrich: I assume these comments are AI generated?! Because it works fine for me using `export type`:

image

I think it doesn't make a difference because this is anyway a `*.d.ts` file. So it cannot export runtime values but types only.

So if you prefer, I can replace the `export type` with simply `export`.

webpro: Indeed the review was generated using AI assistance, and reviewed by me. Here's what I'm seeing:

Image

So indeed seems better to remove `type` and use only `export`.

PR #1331, opened 2026-08-18
Support proxy environment variables in GitLab (resolve #1131)

### LINKED ISSUE #1131: GitLab integration does not support proxy settings
It seems that the GitLab integration does not unterstand standard linux http proxy settings (`http_proxy`, `https_proxy`, `no_proxy` in lower or upper case). I was just building a GitLab CI release job, but the same script that works locally fails in CI with the following error:
```
ERROR Could not authenticate with GitLab using environment variable "GITLAB_TOKEN".
```

I do provide an explicit GITLAB_TOKEN, and this does work locally. As far as I can tell, the GitHub integration has a proxy configuration setting, whereas the GitLab integration does not, which leads me to believe that each integration needs its proxy settings configured explicitly, rather than being pulled from system settings 👎 . Also the `got` Library which it used for GitLab requests has something like this in their own docs regarding proxying:

```
import got from 'got';
import {HttpsProxyAgent} from 'hpagent';

await got('https://sindresorhus.com', {
	agent: {
		https: new HttpsProxyAgent({
			keepAlive: true,
			keepAliveMsecs: 1000,
			maxSockets: 256,
			maxFreeSockets: 256,
			scheduling: 'lifo',
			proxy: 'https://localhost:8080'
		})
	}
});
```
https://github.com/sindresorhus/got/blob/main/documentation/tips.md#proxying

Would it be possibly to implement proxy support across the board, ideally pulling settings from `http_proxy`/`HTTP_PROXY`/etc. and sticking those into octokit, got, etc.?

mrpmohiburrahman:
Resolves #1131.

## Problem

Behind a corporate proxy, a GitLab release fails with a misleading error:

```
ERROR Could not authenticate with GitLab using environment variable "GITLAB_TOKEN"
```

The token is fine. The request never reaches GitLab. The same script works on a laptop and
fails in CI, because the GitLab plugin never reads the standard `http_proxy` / `https_proxy` /
`no_proxy` variables, so the call goes direct and gets dropped — surfacing as an auth failure
rather than a network one. GitHub releases are unaffected: that plugin holds `proxy-agent`,
which reads those variables itself. GitLab had no equivalent.

## Root cause

`lib/plugin/gitlab/GitLab.js` only built a dispatcher when TLS needed customising. With
defaults (`secure: true`, no CA file) `certificateAuthorityOption` stayed `{}` and `fetch` used
undici's global dispatcher, which consults no proxy variables.

Worth noting for anyone wondering why `NODE_USE_ENV_PROXY` isn't the answer: when a CA file or
`secure: false` *is* set, the plugin installs its own dispatcher, and a custom dispatcher
overrides Node's global proxy handling. So the runtime flag cannot help exactly the self-hosted
users most likely to sit behind a proxy. (In the issue thread the supported range was given as
Node 20.19/22.13; `engines` on main is now `^22.21.0 || >=24.0.0`, so the whole supported range
already has the flag — but the shadowing above is what keeps an explicit option correct.)

## Fix

Adds `gitlab.proxy`, **opt-in and defaulting to `false`**, as requested in the thread, so
enabling GitLab releases never silently reroutes traffic. It uses `EnvHttpProxyAgent` from the
existing `undici` dependency — no new dependency.

```js
const needsCustomAgent = Boolean(secure === false || certificateAuthorityFile || proxy);

if (needsCustomAgent) {
  const Dispatcher = proxy ? EnvHttpProxyAgent : Agent;
  this.certificateAuthorityOption.dispatcher = new Dispatcher({ connect: { ... } });
}
```

There is only one `dispatcher` slot, so proxy and TLS have to share it. `EnvHttpProxyAgent`
forwards its `connect` option to both its internal no-proxy `Agent` and its `ProxyAgent`s, so
`gitlab.proxy` composes with `certificateAuthorityFile` and `secure` rather than replacing them.

Scoped to GitLab only, per your note on #1132 about keeping GitLab and GitHub separate.

## Testing

`pnpm test` — 307 passing, up from 303, with the same 4 pre-existing failures before and after
(`git`, `tasks`, `tasks.interactive`; none GitLab-related). `pnpm lint` and `prettier --check`
clean.

Four tests added to `test/gitlab.js`:

- `proxy: true` produces an `EnvHttpProxyAgent` dispatcher
- default config produces no dispatcher at all — proving the option is genuinely opt-in and that
  nothing changes for existing users
- `proxy: true` + `certificateAuthorityFile` produces exactly one dispatcher, and still reads the
  CA file
- an end-to-end test that completes a real TLS handshake against the repo's self-signed HTTPS
  server through `EnvHttpProxyAgent`, trusting only the supplied CA

That last one exists because "the CA still applies when the proxy is on" is the claim most worth
not taking on trust. I checked it bites: mutating the source to drop `connect` under proxy makes
that test — and only that test — fail.

## Notes for reviewers

- **`gitlab.proxy` is a boolean; `github.proxy` is a proxy-URL string.** Same key name, different
  type across sibling plugins. I kept `gitlab.proxy: true` because that is the shape you named in
  the thread, but flagging it: `schema/` is editor tooling only (nothing in `lib/` reads it), so a
  user copying a GitHub config would write `proxy: "http://proxy:8080"`, and that truthy string
  would enable env-var reading while silently ignoring their URL. Happy to rename or accept a URL
  string instead — your call.
- **Adding an option touches four independent lists**, not two: `schema/gitlab.json`,
  `config/release-it.json`, `lib/args.js` and `types/config.d.ts`. The `lib/args.js` entry is
  required — without it the existing test `supports every built-in plugin configuration option`
  fails with `Unknown option '--gitlab.proxy'`.
- **No test drives a live proxy.** Coverage proves the dispatcher is constructed correctly and
  that TLS options survive into it; actual proxy routing is undici's behaviour and is tested
  upstream. Happy to add more if you'd prefer.
- #1132 covered similar ground and was closed for inactivity rather than on design.


lib/plugin/gitlab/GitLab.js
```diff
@@ -16,16 +16,17 @@ class GitLab extends Release {
...
       this.options.certificateAuthorityFile || process.env[certificateAuthorityFileRef] || null;
     this.certificateAuthorityOption = {};
 
-    const needsCustomAgent = Boolean(secure === false || certificateAuthorityFile);
+    const needsCustomAgent = Boolean(secure === false || certificateAuthorityFile || proxy);
 
     if (needsCustomAgent) {
-      this.certificateAuthorityOption.dispatcher = new Agent({
+      const Dispatcher = proxy ? EnvHttpProxyAgent : Agent;
+      this.certificateAuthorityOption.dispatcher = new Dispatcher({
```
webpro, line 29: [P2] Pass GitLab TLS settings through requestTls when proxying

With `gitlab.proxy: true` and `HTTPS_PROXY` pointing to an HTTP CONNECT proxy, both `certificateAuthorityFile` and `secure: false` stop working for a self-signed GitLab server. The request fails certificate verification and authentication reports the same misleading token error. `EnvHttpProxyAgent` forwards the options, but its underlying `ProxyAgent` replaces `connect` and uses `requestTls` for the target TLS connection. Supply the GitLab TLS settings as `requestTls` as well as `connect`, preserving direct and `NO_PROXY` requests. A real local proxy reproduction fails with `DEPTH_ZERO_SELF_SIGNED_CERT` for both configurations and succeeds when `requestTls` is supplied. This matches the [Undici 7.29.0 ProxyAgent API](https://github.com/nodejs/undici/blob/v7.29.0/docs/docs/api/ProxyAgent.md#parameter-proxyagentoptions).

test/gitlab.js
```diff
@@ -468,4 +499,34 @@ describe('GitLab', () => {
...
+    const gitlab = await factory(GitLab, { options });
+    const server = new GitlabTestServer();
+
+    t.after(async () => {
+      await server.stop();
+    });
+
+    await server.run();
+
+    interceptUser(local);
```
webpro, line 527: [P2] Exercise native fetch in the proxy TLS test

The suite installs a global fetch mock, and these interceptors supply both responses without contacting the HTTPS server. Removing the CA configuration still leaves this test passing, with zero requests received by the server. Clearing all proxy variables also means the test cannot exercise TLS through a proxy. Restore `nativeFetch` for this test, remove the response interceptors, and route it through a local CONNECT proxy with a valid test certificate. Assert that the proxy and target receive requests so the test catches lost TLS settings.

webpro (changes requested):
Thanks for adding GitLab proxy support! The TLS settings need to survive proxied requests, and the new TLS test needs to exercise the actual connection. Details inline.

test/gitlab.js
mrpmohiburrahman: Fixed.
The test now uses the real fetch and a real local CONNECT proxy, so requests actually reach the HTTPS server. It checks the proxy and the server both received requests, and it fails now if the CA settings are removed.

lib/plugin/gitlab/GitLab.js
mrpmohiburrahman: Fixed.
The TLS settings are now passed through requestTls as well, so the CA file and secure: false still work when the request goes through a proxy. I verified it with a real local proxy setup.

webpro (approved):
Thanks for the fixes. The TLS settings now survive proxying, and the revised test catches the original regression.

webpro:
Failures in CI not caused by this PR, main's good.

webpro:
:rocket: _This pull request is included in v21.1.0. See [Release 21.1.0](https://github.com/release-it/release-it/releases/tag/21.1.0) for release notes._

PR #1332, opened 2026-08-19
fix(schema): remove fragment from schema $id values

### LINKED ISSUE #1328: jsonschema invalid
I added schema validation for my json files and the validators complain about the `$id` field in the release-it json schema.

`Failed to resolve './git.json' against 'https://unpkg.com/release-it@21/schema/release-it#release-it': base should not have fragment`

https://json-schema.org/understanding-json-schema/basics#declaring-a-unique-identifier and https://json-schema.org/understanding-json-schema/structuring#id recommend using the absolute uri.

Not sure what exactly is correct here though as I have little experience writing them myself.

Edit: draft 04 and 06 seem to validate fine, but 07 (which is defined in the `$schema`) seems to fail.
Edit 2: After some more research this might be an error in the parser, but still not the recommended use of the `$id` if I understand it correctly

DanMat:
Closes #1328

Every `schema/*.json` declares a draft-07 `$id` with a fragment (`release-it#release-it`, `release-it#git`, …). Per draft-07 the root `$id` should be an absolute URI **without a fragment**, and because all five share the base `release-it`, strict validators can't resolve the relative `$ref`s (`./git.json`, etc.) against a fragment-bearing base — the exact error in #1328.

**Fix:** give each schema a distinct, fragment-free filename `$id` (`release-it.json`, `git.json`, `github.json`, `gitlab.json`, `npm.json`). The `./x.json` refs already resolve relative to these.

### Reproduced locally (ajv, draft-07)

Registering the five schemas by `$id` and resolving the root's `./git.json` ref:

**Before (base `main`):**
```
ERROR — can't resolve reference ./git.json from id release-it#release-it
```

**After (this PR):**
```
OK — all 5 schemas registered; release-it root + ./git.json ref resolved
```

I used bare filename ids to keep the relative `$ref`s working without coupling to a host/version; happy to switch to absolute URIs (e.g. `https://unpkg.com/release-it/schema/release-it.json`) if you'd prefer.

*Written with AI assistance; I've reviewed and tested the change and will maintain it.*


lbesecker195 (approved):
Checked with two draft-07 validators. With ajv 8, registering the five schemas by `$id` and compiling the root fails on main with `can't resolve reference ./git.json from id release-it#release-it`; on this branch it compiles, accepts `{ git: { commitMessage }, npm: { publish: false } }` and rejects `{ git: { commit: 'yes' } }`. With @hyperjump/json-schema's draft-07 validator, the schemas registered under their unpkg URLs (`https://unpkg.com/release-it@21/schema/<name>.json`) report even the valid config as invalid on main, and on this branch give valid/invalid as expected, since the fragment-free `$id`s resolve to the same retrieval URLs the `./git.json` refs point at. Nothing in `lib/` or the tests reads `$id` (test/args.js loads the schema files by path). LGTM.

webpro:
Thanks Danny!

webpro:
:rocket: _This pull request is included in v21.1.0. See [Release 21.1.0](https://github.com/release-it/release-it/releases/tag/21.1.0) for release notes._

